PR #1, opened 2019-10-11
Feature: Dashboard

SpencerKaiser:
Adding a sample env file, adding a README for the Slack setup (WIP), adding event and action listeners to support the dashboard

## Pre-Requisites
- [x] Yes, I included and/or modified tests to cover relavent code **OR** my change is non-technical

## Description of Changes

When a user DM's the bot, they'll be presented with a dashboard of ways they can interact with the bot


package.json
```diff
@@ -56,6 +56,7 @@
   },
   "dependencies": {
     "@slack/bolt": "^1.4.0",
+    "@slack/types": "^1.2.1",
```
SpencerKaiser: @johnkahn Not sure what was up, but my IDE was complaining about this being in `devDependencies`. Any ideas?

src/slack/README.md
```diff
@@ -0,0 +1,5 @@
+# Slack App Setup
+- Create an app on the [Slack API Site](api.slack.com)
+- Add a bot user
+- Add event subscriptions for `message.im` (Note: you'll need a local app running with [`ngrok`](https://ngrok.com) or an app url from your deployed app; append `/slack/events` to your URL)
+- Enable _Interactive Components_ and use your app URL and append `slack/events`
```
SpencerKaiser: This file is a WIP; this is just the initial comments I had

src/slack/constants.ts
```diff
@@ -0,0 +1 @@
+export const registerTeamCallbackId = 'registerTeam';
```
SpencerKaiser: @johnkahn I tried using a `d.ts` file here but didn't have any luck... I assume you can't `declare` constants like you can with interfaces?

SpencerKaiser:
@johnkahn our actions are messed up too... not sure why they didn't trigger for this PR

package.json
johnkahn: Probably because they don't follow the standards, smh. Types are supposed to be built into the project or they should exist inside the DefinitelyTyped project imported with `@types/slack` (which doesn't exist)

src/slack/README.md
johnkahn: Works for now

src/slack/actions/index.ts
```diff
@@ -0,0 +1,8 @@
+import { App } from '@slack/bolt';
+import logger from '../../logger';
+import registerTeam from './registerTeam';
+
+export default function register(app: App): void {
+  logger.info('Registering action listeners');
+  registerTeam(app);
```
johnkahn: I'm good with this for now, but in the future, I want to add an auto-import for all files in the folder

src/slack/constants.ts
johnkahn: Yeah you can't, and I was wrong about the standard with types anyway. Because we are inside of a typescript project, we shouldn't use `.d.ts` files. Those are only if you code in Javascript. So for now always use just `.ts` for types/constants

PR #4, opened 2019-10-15
Adding CI/CD Actions

SpencerKaiser:
## Pre-Requisites
- [x] Yes, I included and/or modified tests to cover relavent code **OR** my change is non-technical

## Description of Changes

Adding new actions for CI/CD

### DIFF
--- a/.github/workflows/Deploy.yml
+++ b/.github/workflows/Deploy.yml
@@ -1,0 +1,31 @@
+name: Deploy
+
+on:
+  push:
+    branches: 
+      # Deploy on hack branches (keeps env clean and prevents data loss on deploy)
+      - hack/* 
+
+jobs:
+  release:
+     runs-on: ubuntu-latest
+     steps:
+     - uses: actions/checkout@master
+     - name: login
+       uses: actions/heroku@master
+       env:
+         HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
+       with:
+         args: container:login
+     - name: push
+       uses: actions/heroku@master
+       env:
+         HEROKU_API_KEY: ${{ secrets.HEROKU_API_KEY }}
+       with:
+         args: container:push -a calm-fortress-1234 web
+     - name: release
+       uses: actions/heroku@master
+       env:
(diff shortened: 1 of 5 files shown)


.github/workflows/Deploy.yml
```diff
@@ -0,0 +1,32 @@
+name: Deploy
+
+on:
+  push:
+    branches: 
+      - master
+      # Deploy on hack branches (keeps env clean and prevents data loss on deploy)
+      - hack/*
```
johnkahn: Why `hack/*` branches? I don't fully understand the comment

johnkahn: Will that deploy them to a separate app? Or will that override the prod one?

SpencerKaiser: If we use `master` for our main app, we'll have one app environment and one DB. I guess we can just delete the app afterwards, but my thought was we'd have a new instance of the app for each hack. Want to just stick with `master`?

README.md
```diff
@@ -1,2 +1,5 @@
+![](./workflows?branch=master/Build/badge.svg)
+![](./workflows/Deploy/badge.svg)
```
johnkahn: Wouldn't we want this to be `branch=master` too?

SpencerKaiser: It should default to whatever was last deployed. The other one runs for PRs too, so I want to specify status for the last build on `master`

.github/workflows/Deploy.yml
johnkahn: Ohhh gotcha. No, I like that. That makes it easy to clean it up too, because we won't need to worry about old Slack spaces pointing to an app for another hackathon. I assume they would all have different urls?

README.md
johnkahn: Makes more sense after reading the other comment 👍

.github/workflows/Deploy.yml
SpencerKaiser: That's the plan but I don't want to have to figure out how to derive the app name from the branch name just yet since the only available one has a full ref path associated with it.

PR #8, opened 2019-10-29
[Feature] Setup database

johnkahn:
## Pre-Requisites

## Description of Changes
Created the database entities and migrations


src/index.ts
```diff
@@ -1,3 +1,4 @@
+import 'reflect-metadata';
```
SpencerKaiser: What's this do?

.eslintrc.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
     'import/no-extraneous-dependencies': ['error', { devDependencies: ['**/*.test.ts'] }],
     'max-len': [0],
     'import/prefer-default-export': ['off'],
+    'object-curly-newline': ['off'],
```
SpencerKaiser: Doesn't this open the door to some funky declarations?

src/index.ts
johnkahn: It's for the decorators in the entities. The ✨ magic ✨ behind TypeORM for those beautiful db classes

johnkahn: No clue how it works tbh

.eslintrc.js
johnkahn: The issue was it's erroring on this
```
import { Entity, PrimaryGeneratedColumn, Column, BaseEntity } from 'typeorm';
```
it wants
```
import {
   Entity,
   PrimaryGeneratedColumn,
   Column,
   BaseEntity
} from 'typeorm';
```

johnkahn: Oh wait, it looks like it's just a difference of `prettier` vs `eslint`. Yeah I'll fix that

johnkahn: Ok, so the issue is that `prettier` will break it into multiline based on line length and `eslint` will break it based on the number of properties. So disabling it for `eslint` means `prettier` still catches it. IMO I prefer `prettier`'s method.

PR #42, opened 2019-11-26
Improving test coverage + Fixing an issue where users could register a team when team registration was closed

### LINKED ISSUE #27: [Bug] - Teams can register when teamRegistrationActive is false
### Pre-requisites:

## Description

If a team sends a DM to the bot when `teamRegistrationActive` is true, they'll get a response including the block with the `Register` button. If they click this button after `teamRegistrationActive` is set back to `false`, they'll still be able to register.

## Expected Behavior

When `teamRegistrationActive` is false, teams shouldn't be able to register even if they're able to enter the registration flow. If possible, the original message should be updated to remove the button.

## Steps to Reproduce

- Set `teamRegistrationActive` to true
- DM the bot
- Set `teamRegistrationActive` to false
- Click `Register` and complete the flow

SpencerKaiser:
## Pre-Requisites

## Description of Changes

- Checking the toggle before allowing the dialog to open AND on submit
- Improving the way toggles are retrieved from the DB
- Adding tests to cover config
- Adding tests to cover the test DB
- Adding additional test to cover other files that didn't have 100% coverage

## Related Issues

Fixes #27


SpencerKaiser:
@johnkahn this is ready for review, ignore the one failed status check (see #63 for more details)

src/slack/events/message.ts
```diff
@@ -7,9 +7,10 @@ import logger from '../../logger';
 function register(bolt: App): void {
   bolt.message(async ({ say }) => {
     // TODO: get context (toggles), use to generate specific dashboard blocks
-    const teamRegistrationToggle = (await Config.findOne({ key: 'teamRegistrationActive' })) || { key: 'teamRegistrationActive', value: 'false' };
+    const teamRegistrationActiveKey = 'teamRegistrationActive';
+    const teamRegistrationActive = await Config.findToggleForKey(teamRegistrationActiveKey);
     const context = {
-      [teamRegistrationToggle.key]: teamRegistrationToggle.value === 'true',
+      teamRegistrationActiveKey: teamRegistrationActive,
```
johnkahn: I assume there should be brackets around this? Or should the key actually be `teamRegistrationActiveKey`?

```suggestion
      [teamRegistrationActiveKey]: teamRegistrationActive,
```

src/tests/index.test.ts
```diff
@@ -0,0 +1,70 @@
+import 'jest';
+
+/* eslint-disable @typescript-eslint/no-var-requires, global-require */
+
+const testPort = 3212;
+
+describe('Server', () => {
+  beforeAll(() => {
+    global.console = {
```
johnkahn: Why override `console.log`? Couldn't we just set `LOG_LEVEL=emerg` for tests? Since there are no logs of that level we would hide the logs.

src/tests/index.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    process.env.SLACK_BOT_TOKEN = 'ANOTHER JUNK TOKEN';
+  });
+
+  it('logs which port the server is listening on', async () => {
+    const logger = require('../logger').default;
+    const loggerSpy = jest.spyOn(logger, 'info');
+
+    const server = require('../index').default;
+
+    // Because starting the server is synchronous, we need to wait before seeing if the
```
johnkahn: Do you mean 👇?

```suggestion
    // Because starting the server is asynchronous, we need to wait before seeing if the
```

src/tests/index.test.ts
SpencerKaiser: Good catch, just made that change. I had no idea `console.log` listened to `LOG_LEVEL`, so TIL haha

For what it's worth, this is only here because of the uncontrollable errors thrown by the Bolt SDK when the API key is garbage.

SpencerKaiser:
@johnkahn this is ready for review again

src/tests/index.test.ts
johnkahn: `console.log` doesn't, but we use `winston` around it and it does

SpencerKaiser: The error I was trying to suppress was Slack’s.... soooo maybe they use Winston?

src/tests/index.test.ts
```diff
@@ -55,16 +51,17 @@ describe('Server', () => {
...
 
-  it('listens to defaults to port 3000', () => {
-    delete process.env.PORT;
-    try {
-      const server = require('../index').default;
-      server.close();
-    } catch (err) {
-      // The app is likely already running but we don't need the server anyways
-    }
+  // it('listens on port 3000 as default port', () => {
```
johnkahn: Since we're not running it, can we remove it? (Same with the commented out warning above)

src/tests/index.test.ts
```diff
@@ -1,19 +1,15 @@
...
-    global.console = {
-      ...global.console,
-      log: jest.fn(), // console.log are ignored in tests
-      //   error: jest.fn(),
-      //   warn: jest.fn(),
-      //   info: jest.fn(),
-      //   debug: jest.fn(),
-    };
+    // Hide Bolt API key warnings
+    process.env.LOG_LEVEL = 'emerg';
```
johnkahn: Did this work? Because LOG_LEVEL is only listened to by logging libraries. console.log ignores this. I didn't know the reason for this was because of Bolt.

But if it works, then that means they use a logging library, and it just so happens to work 😆 Which I am fine with

src/tests/index.test.ts
johnkahn: Meant to put this on all the lines, but I guess I only chose the first one. Basically all the commented out stuff not being used.

src/tests/index.test.ts
SpencerKaiser: Sure did! If it ain’t broke...

src/tests/index.test.ts
SpencerKaiser: Since it's in a test file, I'd rather leave it in only because I'd like to get it working eventually and it'd prevent someone from trying to make their own down the road. If you really want to remove it, I can open an issue to track it there but I don't think it has enough priority to warrant an issue haha

PR #53, opened 2019-12-09
Update/judging

SpencerKaiser:
## Pre-Requisites

## Description of Changes

Improving Judging algorithm and testing


src/entities/judgingVote.ts
```diff
@@ -49,92 +49,162 @@ export class JudgingVote extends BaseEntity {
...
+        scores[otherTeam.id] = otherTeamScore;
+
+        // Check to see if all teams have been calibrated
+        const allTeamsAreInCalibration = Object.keys(calibrationCount).length === teams.length;
+        const allTeamsAreCalibrated = !Object.values(calibrationCount).some((value) => value < votesNeededForCalibration);
+        if (allTeamsAreInCalibration && allTeamsAreCalibrated) {
+          calibrationComplete = true;
+          // Update min/max score for traditional scoring
+          const calibratedScores = Object.values(scores);
+          for (let j = 0; j < calibratedScores.length; j += 1) {
```
SpencerKaiser: @johnkahn if I did a `forEach` here, I got some super obscure error: `Function declared in a loop contains unsafe references to variable(s)` which violates the [`no-loop-func` Eslint rule](https://eslint.org/docs/rules/no-loop-func). After reading through the description, I'm still not sure what the issue was but I fixed it just in case  ¯\\_(ツ)\_/¯

SpencerKaiser: Oh and this is necessary to avoid the potential for `minScore` and `maxScore` values being totally wrong in the first few non-calibration passes. No idea why the results weren't wildly broken when this wasn't here because the first few votes judges were always the min/max scores so their score impact was virtually nothing (1st place team vs last team).

src/tests/judging.test.ts
```diff
@@ -193,56 +197,89 @@ describe('score calculation', () => {
...
-    const numJudges = 10;
-    const teams = await createTeamData(numTeams);
-    const judges = await createJudgeData(numJudges);
-
-    const orderedTeams = await visitTeamsAndJudge(judges, teams, 1.0);
-
-    const scores = await JudgingVote.tabulate();
+  it('scoring works as expected without judge volatility and full visitation', async (done) => {
+    // TODO: Achieve 100% in tests with perfect judging
+    const accuracyThreshold = 0.5;
```
SpencerKaiser: Right now our lowest is around 75% accuracy, but very rarely it dips below 70% for some of the smaller tests (5 teams, 5 judges). The overall accuracy is still above 75% so I added a second, stricter test to evaluate the overall leverage accuracy (threshold set below this line, set to 75%)

SpencerKaiser:
@johnkahn @MelindaMalmgren this is ready for review (finally 🙌). 

**TL;DR** - Without modifying the way we select teams for judging (by keeping track of a current score and finding the best match for the "previous" team based on their score), this is likely the best we're going to get. The gist of the algorithm is that ordering of votes has an impact on the way teams are scored (i.e., a team might have 2-3 opportunities to make big points and if *those* votes are scored during calibration, their overall potential plummets) and the order of the teams also has an impact because of the way I select teams for calibration, so we make a number of iterations over a randomized set of teams and a randomized set of votes and then score teams based on their average position across all of those passes.

**CALIBRATION**: Based on the array of teams the method accepts, I take a round-robin approach to grabbing a team with the fewest calibrations and then find a vote from all votes that include them sorted by how many calibrations the _other team_ has. I continue to iterate until all teams have been calibrated. There is still a chance (based on the random selection of teams for votes) that it won't be possible to calibrate all teams without over-calibrating some teams, but the over calibration is done evenly (e.g., if all teams need 2 votes, you might see a couple with 3 calibration votes but none will have 4).

**SCORING**: After calibration, each team is evaluated against their competitor based on their percentile (e.g., Team A has 300 points, Team B has -175 points and the min/max is 1k/-1k respectively; in this situation they're a relatively similar match). Based on who _should_ win mixed with there percentile, the impact to their score may be substantial or minimal with a minimum impact of ±0 and a max impact of ±200.

**TESTING**: I created a suite of tests that use a matrix of `#teams x #judges x visitation%`. I evaluate `accuracy`, which is a measure of the number of errors and the distance of each error (e.g., Team A was supposed to be 1st but they placed 4th; this is much worse than if they had placed 2nd and the accuracy reflects that). Individual tests may have an accuracy of <75% (usually the tests with only a few teams and only a few judges) but on the whole, the accuracy is typically 80% for the full suite. If any individual test falls below 50% accuracy, the suite fails, otherwise the overall accuracy must be above 75% to pass.

PR #69, opened 2020-01-10
Implementing Subscriber Update Slack messaging

### LINKED ISSUE #44: [Feature Request] - Create Slack helper method to send update to subscribers
### Pre-requisites:

A Slack helper method should be created which, when used, will send an update via DM to all subscribed users. The method should be throttled with a reasonable delay to prevent rate limiting issues.

Dependent on the completion of #43

SpencerKaiser:
## Pre-Requisites

## Description of Changes

Implementing initial solution for messaging all subscribers

## Related Issues

- Closes #44 

# TO DO
- [x] Move reusable interfaces to a better place
- [x] Implement UI components to allow users to subscribe
- [x] Update dashboard to allow users to unsubscribe once subscribed
- [x] On subscribe/unsubscribe, modify original dashboard in place
- [x] Implement action auth (waiting on @johnkahn)
- ~[ ] Transition to upsert for subscribe/unsubscribe~


src/api/sendUpdateMessage.ts
```diff
@@ -0,0 +1,30 @@
+import { Request, Response } from 'express';
+import { Subscriber } from '../entities/subscriber';
+import messageUsers from '../slack/utilities/messageUsers';
+import logger from '../logger';
+
+export default async function sendUpdateMessage(req: Request, res: Response): Promise<void> {
+  const { message, secret } = req.body;
+
+  // TODO: Implement auth
+  if (secret !== 'some secret') {
```
johnkahn: For now I would just say use `process.env.ADMIN_SECRET` and we can use that everywhere

johnkahn: Eventually, I'll have this in the config, but that'll work for now

src/slack/utilities/getDashboardContext.ts
```diff
@@ -0,0 +1,16 @@
...
+  [key: string]: boolean;
+}
+
+export async function getDashboardContext(userId: string): Promise<DashboardContext> {
+  const teamRegistrationToggle = (await Config.findOne({ key: 'teamRegistrationActive' })) || { key: 'teamRegistrationActive', value: 'false' };
+  const subscriber = await Subscriber.findOne({ slackId: userId, isActive: true });
+
+  return {
+    [teamRegistrationToggle.key]: teamRegistrationToggle.value === 'true',
+    isSubscribed: Boolean(subscriber),
```
johnkahn: Instead of getting the whole subscriber and casting it to a boolean, why don't we just get the count and then `isSubscribed` could be `count === 1`. It would also be a faster query.

src/slack/utilities/messageUsers.ts
```diff
@@ -0,0 +1,35 @@
...
+        users: users[i],
+      })) as DmOpenResult;
+
+      await slackAPI.chat.postMessage({
+        channel: dm.channel.id,
+        text: message,
+      });
+
+      // Rate limiting for this behavior is capped at ~1 per second, but burst behavior is allowed
+      await waitBeforeNextRequest(500);
```
johnkahn: Do you have a link for this out of curiosity? Just so I can read up on it

src/slack/utilities/messageUsers.ts
```diff
@@ -0,0 +1,35 @@
...
+    } catch (err) {
+      errors.push({ [user]: err });
+    }
+  }
+
+  if (errors.length) {
+    logger.error(`Unable to send update to ${errors.length === users.length ? 'all' : 'some'} users:`, errors);
+    throw errors;
```
johnkahn: We shouldn't throw multiple errors because we won't be able to see the line number or anything. Instead, because we're already logging the errors.

src/slack/utilities/getDashboardContext.ts
```diff
@@ -7,10 +7,10 @@ interface DashboardContext {
 
 export async function getDashboardContext(userId: string): Promise<DashboardContext> {
   const teamRegistrationToggle = (await Config.findOne({ key: 'teamRegistrationActive' })) || { key: 'teamRegistrationActive', value: 'false' };
-  const subscriber = await Subscriber.findOne({ slackId: userId, isActive: true });
+  const isSubscribed = (await Subscriber.count({ slackId: userId, isActive: true })) > 0;
```
johnkahn: For this, can we have just the await? I feel doing async and sync on the same line can get confusing.

PR #72, opened 2020-01-16
Support Request Queue

### LINKED ISSUE #47: [Feature Request] - Create support queue database entities
### Pre-requisites:

An entity called `supportQueueItem` should be created to allow student teams to receive help. Each item should have the following attributes:
- `teamId` (relation to the `team` entity)
- `isActive`: `bool`
- `type`: enum (`IdeaPitch`, `TechHelp`)

### LINKED ISSUE #48: [Feature Request] - Allow teams to request support via Slack
### Pre-requisites:

A team should be able to use Slack to request help from the sponsor. After messaging the bot, the dashboard should include several elements (use the copy below) if the `supportQueueActive` config item is set to `true`:
- _*Pitch your Idea*: come chat with our team and tell us about your idea. We'll give you feedback that'll help you build a better app._
  - Button: `Request Pitch time"
  - Response text: _We'll send your team a DM when we're ready for you to come pitch your idea._
  - Error text: _Looks like your team is already waiting to pitch your idea. We'll send your team a DM when we're ready for you._

- _*Get Tech Help*: Having trouble with something? Come get technical help from our team. `${sponsorSpecialties}`_ (`sponsorSpecialties` should use a comma separated list of languages set as an `ENV` variable, if it exists)
  - Button: `Request Tech Help`
  - Response text: _We'll send your team a DM when we're ready to help you._
  - Error text: _Looks like your team is already waiting to get some help. We'll send your team a DM when we're ready._

Bot listeners for the buttons above should respond with an error if `supportQueueActive` is set to `false`: _Our team isn't available, check back again later!_

### LINKED ISSUE #73: [Feature Request] - API routes for Support Queue
### Pre-requisites:

API routes should be created to allow a user to pop an item off the support queue, close the issue and/or grab a new one.

`closeSupportRequest(id: string)`
`getNextSupportRequest()`

SpencerKaiser:
## Pre-Requisites

## Description of Changes

Adding initial implementation of the Support Queue (`SupportRequest`) entity
- Adding new auth middleware and establishing a better pattern for using it

## Related Issues

- Closes #47 
- Closes #73 
- Closes #48

### TODO
- [x] Add migration


src/api/index.ts
```diff
@@ -75,6 +76,7 @@ api.post('/skip', async (req, res) => {
   res.send(await getJudgeTeams(judge));
 });
 
-api.post('/sendUpdateMessage', sendUpdateMessage);
+api.use('/sendUpdateMessage', sendUpdateMessage);
+api.use('/supportRequest', supportRequestRoutes);
```
johnkahn: You can just add the `adminMiddleware` here to lock down this route, instead of adding it inside of the Router. Makes it easier to see which are locked down and which aren't at a quick glance.

```suggestion
api.use('/supportRequest', adminMiddleware, supportRequestRoutes);
```

src/api/sendUpdateMessage.ts
```diff
@@ -1,15 +1,14 @@
...
+import express from 'express';
+import { adminMiddleware } from './adminMiddleware';
 import { Subscriber } from '../entities/subscriber';
 import messageUsers from '../slack/utilities/messageUsers';
 import logger from '../logger';
 
-export default async function sendUpdateMessage(req: Request, res: Response): Promise<void> {
-  const { message, secret } = req.body;
+export const sendUpdateMessage = express.Router();
+sendUpdateMessage.use(adminMiddleware);
```
johnkahn: I'd like to remove this in favor of the other change in `src/api/index.ts`
```suggestion
```

src/api/index.ts
```diff
@@ -75,6 +76,7 @@ api.post('/skip', async (req, res) => {
   res.send(await getJudgeTeams(judge));
 });
 
-api.post('/sendUpdateMessage', sendUpdateMessage);
+api.use('/sendUpdateMessage', sendUpdateMessage);
```
johnkahn: You can just add the adminMiddleware here to lock down this route, instead of adding it inside of the Router. Makes it easier to see which are locked down and which aren't at a quick glance.

```suggestion
api.use('/sendUpdateMessage', adminMiddleware, sendUpdateMessage);
```

src/api/supportRequest.ts
```diff
@@ -0,0 +1,60 @@
+import express from 'express';
+import { adminMiddleware } from './adminMiddleware';
+import { SupportRequest, SupportRequestStatus, SupportRequestType } from '../entities/supportRequest';
+import logger from '../logger';
+import messageUsers from '../slack/utilities/messageUsers';
+
+export const supportRequestRoutes = express.Router();
+supportRequestRoutes.use(adminMiddleware);
```
johnkahn: Same as others
```suggestion
```

src/api/supportRequest.ts
```diff
@@ -0,0 +1,60 @@
...
+  try {
+    await messageUsers(
+      [nextRequest.slackId],
+      `:tada: We're ready to ${
+        nextRequest.type === SupportRequestType.IdeaPitch ? 'help you with an idea' : 'help with your technical issue'
+      }, so head over to our booth. Feel free to bring other members of your team and make sure to bring your laptop if relevant.`,
+    );
+  } catch (err) {
+    logger.error("Unable to notify users they're support request has been served", err);
+  }
```
johnkahn: Can we wrap this in a `promise` or an `async` function and not `await` it so that the api call isn't waiting for this to resolve before returning?

src/entities/supportRequest.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    for (let i = 0; i < abandonedRequests.length; i += 1) {
+      try {
+        abandonedRequests[i].status = SupportRequestStatus.Abandoned;
+        await abandonedRequests[i].save();
+      } catch (err) {
+        // Ignore coverage for this line
+        /* istanbul ignore next */
+        logger.error('Something went wrong updating stale requests', err);
+      }
```
johnkahn: Can we add this to an array of `promises` and await them all after?

src/entities/supportRequest.ts
```diff
@@ -0,0 +1,126 @@
...
+            status: SupportRequestStatus.InProgress,
+            movedToInProgressAt: new Date(),
+          })
+          .execute();
+
+        await nextRequest.reload();
+        return nextRequest;
+      } finally {
+        retries -= 1;
+      }
```
johnkahn: Not sure I understand this. How could this fail? Because you've already returned `null` if the `nextRequest` doesn't exist, and so reloading will never throw an error.

src/entities/supportRequest.ts
```diff
@@ -0,0 +1,126 @@
...
+  name: string;
+
+  @Column({ nullable: false, default: SupportRequestStatus.Pending, type: 'simple-enum', enum: SupportRequestStatus })
+  status: SupportRequestStatus;
+
+  @Column({ nullable: false, type: 'simple-enum', enum: SupportRequestType })
+  type: SupportRequestType;
+
+  @BeforeInsert()
+  async checkForCurrentRequest(): Promise<void> {
```
johnkahn: Can we rename this to something that is more clear that it fails if there are active requests? Maybe the following?
```suggestion
  async validateNoActiveRequests(): Promise<void> {
```

src/api/supportRequest.ts
SpencerKaiser: The API call is acknowledged up on line 13, this is after the request goes out. The only other thing I could do would be to create a new interface and add the `supportRequest` to it along with a `userNotified` bool. Thoughts? If not, wrapping this in a promise won't change anything.

src/entities/supportRequest.ts
SpencerKaiser: I thought about that, but A) I'd rather not bomb the DB with a bunch of simultaneous calls, and B) there should only ever be one here so it won't change much.

src/entities/supportRequest.ts
SpencerKaiser: Two requests exist: 1 and 2
User 1 and user 2 `getNext` at the same time and both get request 1, one of the two will successfully update the record, the other will not. The one who fails will retry and get 2. If there are no records, it just bails entirely.

src/api/supportRequest.ts
johnkahn: AFAIK it doesn’t actually send the response until it returns though

src/entities/supportRequest.ts
johnkahn: Fair point

src/api/supportRequest.ts
SpencerKaiser: Might as well make the interface then. Cool?

src/entities/supportRequest.ts
johnkahn: But what is actually failing? The update won't fail because it'll just update 0 records, and the reload won't fail because that has the record. Basically they would both get the same one.

Also, what happens when all the retries go through? It looks like it doesn't actually return anything if it has to use all retries

src/entities/supportRequest.ts
```diff
@@ -18,6 +18,18 @@ export enum SupportRequestErrors {
...
+  const prefix = Math.random()
+    .toString(36)
+    .substring(2, 15);
+
+  const suffix = Math.random()
+    .toString(36)
+    .substring(2, 15);
+
+  return prefix + suffix;
+};
```
johnkahn: Could we refactor this into a util file? Looks like `team.ts` uses the same function

PR #88, opened 2020-01-23
Updating judging to return an actual score

### LINKED ISSUE #87: [Feature Request] - Create functionality for bonus points to be awarded for teams who pitched
### Pre-requisites:

A team who pitches should be awarded with an extra bit of points for getting feedback and trying to build an even better app. The changes should be made to the API endpoint which should modify the original object and return a new object with additional fields like `bonusPointsAwarded`, `totalScoreWithBonusPoints`, `numberOfPitches`, `numberOfTechHelpSessions`.

SpencerKaiser:
## Pre-Requisites

## Description of Changes

- Updating scoring to return an actual score instead of an average position
- Updating scoring to include the option for a bonus score

## Related Issues

- Closes #87

### DIFF
--- a/src/entities/judgingVote.ts
+++ b/src/entities/judgingVote.ts
@@ -65,18 +65,32 @@ export class JudgingVote extends BaseEntity {
     }
 
     // Average scores from all passes
-    const normalizedScores: { [id: number]: number } = {};
+    const normalizedScores: { [id: string]: number[] } = {};
     Object.values(initialScores).forEach((scores) => {
       Object.values(scores).forEach((teamScore) => {
-        normalizedScores[teamScore.id] = (normalizedScores[teamScore.id] || 0) + teamScore.score;
+        normalizedScores[teamScore.id] = normalizedScores[teamScore.id] ? [...normalizedScores[teamScore.id], teamScore.score] : [teamScore.score];
       });
     });
 
+    const trimmedMeanScores: { [id: string]: number } = {};
+    const percentOfOutliersToRemove = 0.2;
+    const outliersToRemoveFromEachSide = Math.max(Math.floor(percentOfOutliersToRemove * randomJudgingIterations) / 2, 1);
     for (let i = 0; i < Object.keys(normalizedScores).length; i += 1) {
-      normalizedScores[i] /= randomJudgingIterations;
+      const teamId = Object.keys(normalizedScores)[i];
+      const sortedScores = normalizedScores[teamId].sort();
+      // Remove outlier(s) from front
+      sortedScores.splice(0, outliersToRemoveFromEachSide);
+      // Remove outlier(s) from end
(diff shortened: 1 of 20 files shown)


src/tests/api/judging.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    const response = await supertest(app)
+      .get('/api/judging/results/')
+      .set({
+        Authorization: adminSecret,
+      })
+      .expect(200);
+
+    const results: ExpandedTeamResult[] = response.body;
+    expect(results).toBeDefined();
+  });
```
johnkahn: Isn't this just testing that something got returned? Not that it actually got expanded judging results, since Typescript gets removed at compilation?

src/tests/entities/judging.test.ts
```diff
@@ -1,13 +1,12 @@
 import 'jest';
-import shuffle from 'shuffle-array';
-import { createDbConnection, closedbConnection } from './testdb';
-import { Team } from '../entities/team';
-import { Judge } from '../entities/judge';
-import { JudgingVote, insufficientVoteCountError } from '../entities/judgingVote';
-import logger from '../logger';
-import { createJudgeData, createTeamData } from './utilities';
+import { createDbConnection, closedbConnection } from '../testdb';
```
johnkahn: I tried to ignore it, I swear, but can we capitalize the `d`?

```suggestion
import { createDbConnection, closeDbConnection } from '../testdb';
```

src/tests/api/judging.test.ts
johnkahn: Isn't this just testing that something got returned? Not that it actually got expanded judging results, since Typescript gets removed at compilation?

src/tests/entities/judging.test.ts
johnkahn: Also just realized that it was like this before, meaning I approved this before

Soccer player holding his head, with the text what have I done

src/tests/api/judging.test.ts
SpencerKaiser: That’s exactly what’s going on... is there another way to test it?? I couldn’t find anything as casting will succeed since this interface extends the other

src/tests/entities/judging.test.ts
SpencerKaiser: Yea good catch, I’ll fix that

src/tests/api/judging.test.ts
johnkahn: I meant shouldn’t we test what is actually returned? Like do an `.isEqual` to the object we expect?

src/entities/judgingVote.ts
```diff
@@ -128,6 +123,10 @@ export class JudgingVote extends BaseEntity {
     const remainingVotes = [...votes];
 
     // Iterate over all votes
+    const watScores: {
```
johnkahn: wat?

PR #96, opened 2020-01-25
Show recent pop and add no show button

johnkahn:
## Pre-Requisites

## Description of Changes

Added the no show button and it now shows recent pop when you get next in queue


.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     'object-curly-newline': ['off'],
     '@typescript-eslint/no-use-before-define': ['off'],
     'no-await-in-loop': ['off'],
+    'operator-linebreak': ['off'],
```
SpencerKaiser: Can we move this to an exception where needed? I like this rule 😬

Issue #143, opened 2020-09-10
[Bug] - Documentation doesn't specify required Node version


SpencerKaiser:
### Pre-requisites:
- [x] I looked through both [open and closed issues](../issues?utf8=✓&q=is%3Aissue) and did not find an existing bug.

## Description

We need to specify that version 12.x of Node.js is required for the app and that 13+ is not supported

## Expected Behavior

Docs will clearly specify what version of Node is supported

## Actual Behavior

No docs are present

SpencerKaiser:
@oscargoddard dangit, I missed the related issue for #144. That PR description didn't include that special `Closes` syntax, so this issue is open despite the fact that the PR fixing it was merged.

Next time make sure to add the following so the issue is automatically closed on merge:
>Closes #143

PR #144, opened 2020-09-11
Update to Readme.md

oscargoddard:
[Bug] - Documentation doesn't specify required Node version #143
added the necessary parts to let users know NODE version 12.x is needed.

## Pre-Requisites

## Description of Changes

specified that version 12.x of Node.js is required for the app and that 13+ is not supported


johnkahn (approved):

README.md
```diff
@@ -41,6 +41,9 @@ When the app is deployed to a cloud environment, the `DATABASE_URL` `.env` var w
 During the installation process (if you follow the steps on [postgresql.org](https://www.postgresql.org/download/)), you will be prompted to set a password - make sure to use something you'll remember.
 
 ## Starting the App
+
+If NODE is not already installed, to use the Hangar you first need to download it from [here](https://nodejs.org/en/) Be sure to use the version v12.18.3 LTS. Hangar now only supports v12.x.
```
SpencerKaiser, line 45: Can you move all of this to a new section right below `# Local Development`? The section should be `## Node` and it should contain the revised text below:
```suggestion
Hangar uses Node.js as its runtime. To use the Hangar, you'll need to [download](https://nodejs.org/en/) and install the LTS (Long Term Support) version, which is currently 12.x. If you decide to use another version of Node such as 13.x or 14.x, you may experience unexpected issues.
```

For the future, avoid using specific versions in documentation (e.g., "v12.18.3") because you never know what could happen to the project. For example, it's not unheard of for projects to outright pull versions because of critical vulnerabilities (i.e., deprecating and pulling the binary for v12.18.3 so it's virtually impossible to install that version).

SpencerKaiser (changes requested):
A bit of feedback before we merge it

README.md
SpencerKaiser: @oscargoddard make sure to review the other comment above about the new section

oscargoddard (commented):
I added ## Node under # Local Development

README.md
oscargoddard: I moved them under ## Node

README.md
```diff
@@ -30,6 +30,10 @@ This app is still in the early stages of development and only some functionality
 ---
 
 # Local Development
+
+##  Node
```
SpencerKaiser, line 34:
```suggestion
## Node
```

SpencerKaiser (changes requested):
Last correction 🙂

README.md
oscargoddard: Committed the Suggestion. They never teach things like this in school. Thats what we call real life experience. Thank you for the feedback

SpencerKaiser (approved):
This is great, thanks @oscargoddard!

PR #171, opened 2020-09-24
 Default config items should be present on initial app start 

### LINKED ISSUE #89: [Feature Request] - Default config items should be present on initial app start
### Pre-requisites:

The app should first start with a default set of config items, which should be set in the DB if they don't exist already. We should utilize the seed functionality of Typeorm to make this happen: https://medium.com/@bansalsushil_34403/how-to-seed-typeorm-d9637a5948cc

smitshah2531:
## Pre-Requisites

## Description of Changes

Added new migration for default config items

## Related Issues

This PR will close #89 - Default config iteams should be present on initial app start


package.json
```diff
@@ -90,7 +90,6 @@
     "@types/express": "^4.17.1",
     "@types/jest": "^24.0.20",
     "@types/luxon": "^1.21.0",
-    "@types/node": "^11.15.0",
```
johnkahn, line 93: Why was this removed? I believe it's still required

src/migration/1600969675208-defaultConfigItems.ts
```diff
@@ -0,0 +1,23 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class defaultConfigItems1600969675208 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.query(
+            'INSERT INTO "configuration" ("key", "value") VALUES ("teamRegistrationActive", "true")', undefined
```
johnkahn, line 7: @SpencerKaiser thoughts on if these should be true or false?

johnkahn (changes requested):

package.json
smitshah2531: It was removed while I was trying to installing typeorm. I will install the `@types/node` and fix it!

src/migration/1600969675208-defaultConfigItems.ts
smitshah2531: Ya. I was confused about this too. I guess @SpencerKaiser mentioned we need to have default values disabled. So should I change it to false?

package.json
smitshah2531: @johnkahn do we want to keep version "11.15.0"? I installed it and it is showing me version "14.11.2"

package.json
```diff
@@ -65,6 +65,7 @@
     "@slack/types": "^1.9.0",
     "@slack/web-api": "^5.12.0",
     "@slack/webhook": "^5.0.2",
+    "@types/node": "^14.11.2",
```
SpencerKaiser, line 68: Revert this change; this belongs in the `devDependencies` section
```suggestion
```

package.json
SpencerKaiser: @johnkahn it was moved to `dependencies`

@smitshah2531 unless there was a reason for the version bump, let's handle that in a separate issue. Just revert this change back to the original.

SpencerKaiser (changes requested):
Two quick changes and then we can merge!

package.json
SpencerKaiser, line 93: You need to add this dependency back. Also make sure to fix the associated changes in `package-lock.json` (just revert back to what they were.

The end state is that the `Files changed` tab here shouldn't include either `package.json` or `package-lock.json`

package.json
```diff
@@ -117,4 +117,4 @@
     "ts-jest": "^24.0.0",
     "ts-node": "^8.4.1"
   }
-}
+}
```
SpencerKaiser, line 120:
```suggestion
}

```

package-lock.json
```diff
@@ -15953,4 +15953,4 @@
       }
     }
   }
-}
+}
```
SpencerKaiser, line 15956:
```suggestion
}

```

SpencerKaiser (approved):
🚢

PR #173, opened 2020-09-25
Adding the /getAll route as well the status query parameter

### LINKED ISSUE #158: [Feature Request] - Create API route to return all support items
### Pre-requisites:

We need a new endpoint to allow an admin to retrieve all support items from the UI

Create a new route nested within `api/supportRequest/` for `getAll`, which should return a list of all support request items. The method should default to only returning `Pending` items, however, a query parameter of `status` should be utilized to provide others if needed (list of values in `entities/supportRequest.ts`).

Ulfingrad:
## Pre-Requisites

## Description of Changes

@msiddique933 and I Added a route called `getAll` which returns all support requests in the database. Additionally there is the option for a `status` parameter which will allow the user to fetch support requests based on their status. Notably we did not add testing, so that is something that needs to be implemented in the future.

## Related Issues

Closes #158 which was a feature request for adding a `/getAll` route for support requests. This includes `status` query parameter in which a user can search the support requests by their status.


Ulfingrad:
Did I mess up when it comes to bypassing testing? I used `--no-verify` when committing. Was I supposed to put it somewhere else or something?

johnkahn:
No verify will only skip tests on your local machine, but it'll still run it on github and won't let us merge it, so tests will need to be added for it to be mergable

@SpencerKaiser

SpencerKaiser:
@johnkahn yea we chatted about it yesterday. I didn't consider the added difficulty of learning to write tests for this issue so I thought it was fair to have them focus on it later

src/api/supportRequest.ts
```diff
@@ -12,6 +12,34 @@ export interface NextSupportRequestResponse {
...
+        requests = await SupportRequest.find({ status: SupportRequestStatus.Complete });
+        res.send(requests);
+      case "Pending":
+        requests = await SupportRequest.find({ status: SupportRequestStatus.Pending });
+        res.send(requests);
+      case "InProgress":
+        requests = await SupportRequest.find({ status: SupportRequestStatus.InProgress });
+        res.send(requests);
+      default:
+        res.send("It doesn't look like that's a valid status!");
```
SpencerKaiser, line 35: It's super important with situations like this that you follow HTTP standards to communicate with the front end, so we need to update this to send a status back to the front end. In this situation, the frontend goofed up in making the request, so the status should indicate that the client was at fault. I'll let you figure out what status to use. I think the correct syntax is `res.status(XXX).send('Invalid status')`

SpencerKaiser (changes requested):
A couple of quick things, but this is a great start!!

SpencerKaiser:
Also @Ulfingrad make sure your issue description is human-readable. Go take a look at other issues for a better idea of what we're looking for, but for this I'd suggest "Add new /getAll route for support requests" or something similar.

SpencerKaiser (approved):
:shipit:

PR #180, opened 2020-09-26
Removing dependency on next app iniitlization while not in prod

SpencerKaiser:
## Pre-Requisites

## Description of Changes

- Removing dependency on next app initialization when NOT in production to speed up local app restarts
- Fixing an issue where app start logs were displayed despite a mocked logger
- Fixed previously broken Slack initialization tests

### DIFF
--- a/src/tests/app.test.ts
+++ b/src/tests/app.test.ts
@@ -2,7 +2,7 @@ import 'jest';
 import supertest from 'supertest';
 import logger from '../logger';
 
-const loggerErrorSpy = jest.spyOn(logger, 'error').mockImplementation();
+jest.spyOn(logger, 'error').mockImplementation();
 const loggerInfoSpy = jest.spyOn(logger, 'info').mockImplementation();
 
 // eslint-disable-next-line import/first
@@ -42,11 +42,10 @@ describe('app', () => {
     expect(processExitSpy).toBeCalledTimes(1);
   });
 
-  it('fails to initialize Slack, logs an error, but continues in test env', async () => {
+  it('will not exit the process when Slack tokens are not provided and the NODE_ENV is "test"', async () => {
     Object.defineProperty(process.env, 'NODE_ENV', { value: 'test' });
     jest.resetModules();
     await initSlack();
-    expect(loggerErrorSpy).toBeCalledTimes(1);
     expect(processExitSpy).not.toBeCalled();
   });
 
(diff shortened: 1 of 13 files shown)


johnkahn (approved):

src/tests/app.test.ts
```diff
@@ -1,32 +1,75 @@
...
   it('responds successfully to GET /', (done) => {
-    const { app } = require('../app');
     supertest(app)
       .get('/')
       .expect(200, done);
   });
 
-  it("fails to start when Slack tokens are not provided and env isn't test", () => {
-    delete process.env.SLACK_BOT_TOKEN;
+  it("fails to initialize Slack when Slack tokens are not provided and env isn't test and exits", async () => {
```
johnkahn, line 38: Wanted to move this around so that the intent was more clear

```suggestion
  it("will exit the process when Slack tokens are not provided and NODE_ENV !== "test", because it cannot initialize Slack", async () => {
```

src/tests/app.test.ts
```diff
@@ -1,32 +1,75 @@
...
     Object.defineProperty(process.env, 'NODE_ENV', { value: 'development' });
-    expect(require('../app').app).toThrow();
-    // Reset the env variable to 'test'
+    jest.resetModules();
+    await initSlack();
+    expect(processExitSpy).toBeCalledTimes(1);
+  });
+
+  it('fails to initialize Slack, logs an error, but continues in test env', async () => {
```
johnkahn, line 45: To match previous change request.

```suggestion
  it('will not exit the process when Slack tokens are not provided and NODE_ENV === "test", but will log an error', async () => {
```

What's the purpose of it logging an error in the test environment if it's mocked? I would think it would be better for the error to be logged when it quits, otherwise it's not providing much use. And if it is currently being logged when it quits, then I would just move
```typescript
expect(loggerErrorSpy).toBeCalledTimes(1);
```
up into the above test so that we're testing what we want to happen.

johnkahn (changes requested):

src/tests/app.test.ts
SpencerKaiser: The purpose of this is to make sure tests still function correctly when they aren't specifying Slack credentials. I agree the logger serves no purpose; I'll remove it

johnkahn (approved):

PR #182, opened 2020-09-30
Fix connection check

johnkahn:
## Pre-Requisites

## Description of Changes
Fixed a bug where an error was thrown on startup because `getConnection()` throws an error instead of returning `null`

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -35,18 +35,14 @@ app.get(
 app.use('/api', apiApp);
 
 async function initDatabase(): Promise<void> {
-  // hasConnection is used in tests that are testing the initialization of the app, so NODE_ENV will not be 'development' or 'production'.
+  // This is needed for tests that are testing the initialization of the app.
   // The tests already connect to an in-memory SQLite database, so connecting to Postgres would cause those tests to fail.
-  let hasConnection;
-
   try {
     getConnection();
-    hasConnection = true;
-  } catch (err) {
-    hasConnection = false;
-  }
+    return;
+  } catch (err) {} // eslint-disable-line no-empty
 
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -35,7 +35,15 @@ app.get(
...
 
 async function initDatabase(): Promise<void> {
-  if (!getConnection() && process.env.NODE_ENV !== 'test') {
+  let hasConnection = false;
+  
+  try {
+    getConnection();
+    hasConnection = true;
+  } catch(err) {
+  }
```
SpencerKaiser, line 44:
```suggestion
      // Ignore this; we'll just create a connection now
    }
```

SpencerKaiser (changes requested):
Let's add a comment in the catch block for clarity

@johnkahn would reverting this back to a const fix the issue?

`const initDatabase = async (): Promise<void> => {`

src/app.ts
johnkahn: Reverting back to a const wouldn't fix it unfortunately. It's the same issue we had with all the mocking before. I'll add the comment though to explain the purpose of it.

src/app.ts
```diff
@@ -35,7 +35,16 @@ app.get(
...
+  let hasConnection;
+
+  try {
+    getConnection();
+    hasConnection = true;
+  } catch (err) {
+    hasConnection = false;
+  }
+
+  if (!hasConnection && process.env.NODE_ENV !== 'test') {
```
SpencerKaiser, line 47: Instead of using this bool, could we just return from the success block of the try/catch?

johnkahn: Yeah, that's better lol

SpencerKaiser (approved):

PR #188, opened 2020-10-10
Adds get specific endpoint

### LINKED ISSUE #179: [Feature Request] - Create endpoint for handling a support request
### Pre-requisites:

Our frontend user should be able to serve a _specific_ support request instead of getting the next of any type.

Create a new api endpoint for `PATCH` - `api/supportRequest/` which accepts a supportRequest id as part of the body and then marks that support request as being `inProgress`. This functionality should mirror what is taking place in `getNext` but should operate on a specific support request.

msiddique933:
Updating code on my github to include all the code written, This shoud fix the 179 issue as well

## Pre-Requisites

## Description of Changes
Worked with Austin to add the ability to get a specific support request by ID. 

## Related Issues

 Closes #179 - Get specific endpoint


Ulfingrad:
We are unsure as to why the linting is angry. Can white space set it off? I don't know if I'm using Eslint correctly, but from what I could tell it didn't have any issues (which it should have based on it failing here).

Ulfingrad:
Ok, I looked into it a bit more and it turns out I didn't have Eslint set up properly (me dumb). With it now working, there seems to indeed be some issues with the whitespace that we have set up. If I am correct, then it should work with lines 163, 184, 201, and 218 deleted or shortened in the test file (trailing spaces are not allowed, but they are included on those lines). One final note: Eslint was not happy with any line endings on my machine which is due to the expectation of LF instead of CRLF. I tried replacing the value of end_of_line to be CRLF instead of LF in .editorconfig, but Eslint didn't seem to care. I assume these errors should be ignored as they are on literally every line, but is there a way to disable this error altogether that I haven't tried?

johnkahn (changes requested):
@Ulfingrad For the line endings, you'll need to update the settings of your editor to use LF line endings. In VS Code the setting is called "eol", but our pre-commit hooks automatically convert the line endings to LF, which is why we don't see them in the GitHub check output.

And then looks like you may want to run prettier in the following files. The easiest way to do that is to install the `Prettier` extension in Visual Studio Code, open the file, `ctrl + shift + p` and type `Format Document`.

Files to run prettier on:
`src/api/supportRequest.ts`
`src/tests/api/supportRequest.test.ts`

johnkahn (approved):

PR #223, opened 2020-11-06
support request page

### LINKED ISSUE #196: [Feature Request] - Create Support Request Page
### Pre-requisites:

We need to make the support experience even cleaner and make the UI more functional

The existing support request UI items should be moved to a new `/support/[secret]` route which will allow users to navigate to a specific support-request page. The secret in the URL should be pulled from the environment variable `SUPPORT_SECRET`

Note: You'll need to move all the user name field (at the top of the page) and logic over as well

oscargoddard:
## Pre-Requisites

## Description of Changes

created src/pages/support/[secret].tsx and moved required functionality. 

## Related Issues

Closes #196


oscargoddard:
Everything is fixed. It is not without without problems. There is only one issue. One of the tests is failing. it expects In Progress gets pending.

SpencerKaiser (approved):

src/pages/index.tsx
```diff
@@ -1,15 +1,7 @@
+/* eslint-disable react/jsx-key */
```
johnkahn, line 1: We shouldn't disable this rule, we need the keys to be set so that React can optimize our lists.
```suggestion
```

src/pages/index.tsx
```diff
@@ -201,66 +73,22 @@ const AdminPage: NextComponentType = () => {
...
-                  (lastPop.supportRequest !== null) ?  
-                    (lastPop.userNotified
-                      ? `${lastPop.supportRequest.name} has been notified to come over`
-                      : `There was a problem notifying ${lastPop.supportRequest.name}, send them a message and tell them you're ready for them`)
-                    : 'It appears you are trying to access a support request that does not exist!'}
+              <h2 className="font-weight-normal">Config Items</h2>
+              {results.length === 0 && <div className="alert alert-info mt-3">No config items to display 🤔</div>}
+
+              {configItems.map((configItem) => (
+                <div className="form-group ml-sm-3 mr-3 mb-2">
```
johnkahn, line 85: Missing the key on here, which was probably what the rule was complaining about. A key is just a value that is unique from the other items in the loop so React can know which is which when it's rendering an update. It's also best to not make it the index, because otherwise if the order changes it won't know if the order changed or if the contents of the item changed.
```suggestion
                <div key={configItem.key} className="form-group ml-sm-3 mr-3 mb-2">
```

johnkahn (changes requested):
A couple quick changes

johnkahn (approved):

PR #228, opened 2020-11-30
Implements the registerTeam command in discord

### LINKED ISSUE #210: [Feature Request] - Implementation of Team Formation in Discord
### Pre-requisites:

## Is your feature request related to a problem or an open issue? Describe the problem, and if applicable, reference the issue.

We need to add the functionality of team formation in discord. 

The user should be able to generate a team during the hackathon. 

This issue relies on the research on team formation #209

## Additional context

I will be working with @msiddique933 

@msiddique933 and @Peruggia29 to implement this

Ulfingrad:
## Pre-Requisites

## Description of Changes

Worked with @msiddique933 and @smitshah2531 to implement team formation in Discord. Below is a link to showcase what it looks like:

https://www.youtube.com/watch?v=vwDpp6SrmcU&ab_channel=AustinSlobodnik

We have not done testing as testing is going to be quite beastly for this section, but the command itself should be fully implemented for the most part! I did not showcase it in the video, but this process is able to handle multiple users at once without messing with each other's input.

## Related Issues

Closes #210


Ulfingrad:
It seems for testing we will need a way to mock out time passage as well as a way to mock out the whole process (sequential calls to registerTeam doesn't do the trick).

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+
+const promptArr = [
+  'Who are you hacking with? You have 5 minutes to send a reply. If you are hacking alone, simply write `none`!\n**Be sure to input their Discord ID in a comma-separated list e.g. 111222333444555,666777888999000,...**',
+  "What's your team/app name? You have 2 minutes to answer.",
+  "What's your table number (e.g. 42)? You have 2 minutes to answer.",
+  'What does your project do? How will it make a difference? What technologies are used? You have 5 minutes to answer.',
+];
+const componentArr = ['members', 'name', 'table', 'description'];
+const timeArr = [300000, 120000, 120000, 300000];
+const currentUsers: string[] = [];
```
SpencerKaiser, line 57: I see what you're doing here with this value, however, what you're doing is making the application `stateful` in the sense that this file has a chunk of data held in memory that is very important to the user experience. To highly the impact of being stateful, follow this example:
1. User starts into the `!registerTeam` flow
1. Their id is added to `currentUsers`
1. They continue through the necessary fields until they get to description
1. The app restarts (either due to an error in the runtime environment itself or because of a deployment)
1. The user submits their description
1. The app has no knowledge of this user being in the flow and their data is lost

We need to rely on the `discordContext` table and use that to store whether the user is in the flow or not and, if they are, what values they've provided us with so far.

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  addTable(t: number): void {
+    this.table = t;
+  }
+}
+
+let registrationInProgress = false;
+let decrement = false;
+let findName: Team;
+let findTable: Team;
```
SpencerKaiser, line 47: Because all these values exist at the root scope of this file, they'll be overridden as other users enter into this flow. For example, if 5 users use `!registerTeam` simultaneously, these values will bounce around and lose their correctness. 

We need to create a map of these items to a given user's id to track them through this flow.

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
+import Discord from 'discord.js';
+
+import { Config } from '../../../entities/config';
+import logger from '../../../logger';
+import { colors } from '../../constants';
+import { Team } from '../../../entities/team';
+
+class TeamInfo {
```
SpencerKaiser, line 8: An interface should suffice for this data type. We can use a class but it'll just create overhead and require us to write more tests with little added value.

src/discord/events/message/index.ts
```diff
@@ -25,39 +26,47 @@ export const commands: Command[] = [
...
+    trigger: '!registerTeam',
+    description: 'Leads the user through the process of registering a team',
+    handler: registerTeam,
+  },
 ];
 
 export async function message(msg: Discord.Message): Promise<void> {
-  // Make sure the bot doesn't respond to itself
-  if (msg.author.id === client.user.id) return;
+  if (!getUsers().includes(msg.author.id)) {
```
SpencerKaiser, line 38: Remove; see comments below

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+let registrationInProgress = false;
+let decrement = false;
+let findName: Team;
+let findTable: Team;
+
+const promptArr = [
+  'Who are you hacking with? You have 5 minutes to send a reply. If you are hacking alone, simply write `none`!\n**Be sure to input their Discord ID in a comma-separated list e.g. 111222333444555,666777888999000,...**',
```
SpencerKaiser, line 50: What was the intent behind time-boxing their reply? I don't think it really adds value for us and it causes us to have to write more tests (and tests that involve time are **very** unpleasant)

Was there a reason we have to use ID's instead of having the user @mention others? If they can @mention others, we should use that approach because asking them to find ID numbers will add a bunch of complexity and inevitably lead to questions...

```suggestion
  'Who are you hacking with? (Reply by @mentioning each of your team mates here. If you're solo, reply with `none`',
```

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+  "What's your team/app name? You have 2 minutes to answer.",
```
SpencerKaiser, line 51: Let's clarify this; we've had questions in the past about whether we needed either/or or if we needed two different values
```suggestion
  "What's your team or app name?",
```

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+async function question(
+  m: Discord.Message,
+  f: (m: Discord.Message) => boolean,
+  prompt: string,
+  t: number,
+  component: string,
+  currentTeam: TeamInfo,
+): Promise<void> {
+  await m.author.send(prompt);
+  await m.channel
```
SpencerKaiser, line 68: Ah... now I see why you had the time above. @johnkahn thoughts on this pattern? I think we should avoid it as it creates complexity for our tests and doesn't necessarily benefit the user experience.

My guess is that Discord created this to make it easier for others to create apps, but I think our `discordContext` pattern is stronger.

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+      max: 1,
+      time: t,
+    })
+    .then(async (collected) => {
+      const input = collected.first().content;
+      if (input === '!quit') {
+        registrationInProgress = false;
+        m.author.send('You successfully halted the signup process! Feel free to keep using the bot like normal.');
+      } else {
+        switch (component) {
```
SpencerKaiser, line 79: Assuming @johnkahn is on board with us refactoring to avoid the listener pattern, we should store a variable (`step` maybe?) in our `discordContext` for the user and simply use that to figure out how to react.

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+        }
+      }
+      if (registrationInProgress) {
+        await question(msg, filter, promptArr[promptArr.length - 1], timeArr[timeArr.length - 1], componentArr[componentArr.length - 1], team).then(
+          () => {
+            const newTeam = new Team(team.name, team.table, team.description, team.idArr);
+            msg.author.send({
+              embed: {
+                color: colors.info,
+                title: '**You are signed up :partying_face:**',
```
SpencerKaiser, line 178:
```suggestion
                title: '**You're signed up :partying_face:**',
```

src/discord/events/message/registerTeam.ts
```diff
@@ -0,0 +1,214 @@
...
+                description: 'Below is what has been submitted as your team!',
```
SpencerKaiser, line 179:
```suggestion
                description: 'Here are the details for your team:',
```

SpencerKaiser (changes requested):
This is a great first start, but we're going to have some issues with multiple users entering this flow simultaneously.

We should also refactor to utilize `DiscordContext` to track our user during the flow and allow us to abandon the timeouts for commands.

Although not directly related to this flow, we should add a new `!exit` listener within the general message listener to clear the context for a given user if needed. When the bot starts the flow, this keyword should be mentioned to allow the user to escape the flow if needed.

Let us know if you'd like to work through some of this on a call tomorrow.

SpencerKaiser (changes requested):

SpencerKaiser (approved):

PR #229, opened 2020-12-01
UI component for Editing config items

### LINKED ISSUE #204: [Feature Request] - UI Component for Editing Config Items
### Pre-requisites:

Currently there isn't an ability to change config items on the front-end side of things.

As it stands, the user has to go into the database in order to change the config items. We want to implement a way for the user to interact with and change these config items on a page.

This implementation can be added in a few ways. One idea is to hook each config item to a button and have the button serve as a toggle (we will also have a clear way to show which toggle option it is set to). Another way is to have a drop-down menu for each config item. There are multiple ways we could do this, so these are simply idea pitches. Additionally, we need to figure out which page this should be added to (admin page? Its own page? etc.).

oscargoddard:

## Pre-Requisites

## Description of Changes

created a button at the admin page
created a function to toggle the key

## Related Issues

closes #204

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -28,7 +28,6 @@ const AdminPage: NextComponentType = () => {
     promises.push(
       (async (): Promise<void> => {
         const res = await fetch('/api/config');
-
         if (!res.ok) {
           setError(true);
           return;
@@ -62,6 +61,16 @@ const AdminPage: NextComponentType = () => {
     Promise.all(promises).then(() => setLoading(false));
   });
 
+  const handleChange = () => async (): Promise<void> => {
+    await fetch('/api/config', {
+      method: 'POST',
+      body: JSON.stringify({}),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  };
+
   if (loading) return null;
 
   if (error) return <h4>Error loading, check the console</h4>;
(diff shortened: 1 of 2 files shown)


src/pages/index.tsx
```diff
@@ -86,6 +95,9 @@ const AdminPage: NextComponentType = () => {
                     {configItem.key}
                   </label>
                   <input type="text" className="form-control" id={configItem.key} placeholder="Team name" value={configItem.value} disabled />
+                  <button type="button" className="btn btn-primary mb-2 mt-3" onClick={handleChange}>
+                    Change
+                  </button>
```
johnkahn, line 100: This is where you'll need to pass the key and new value to your `handleChange` function. I would also change the button text to reflect exactly what it's doing.

```suggestion
                  <button type="button" className="btn btn-primary mb-2 mt-3" onClick={() => handleChange(configItem)}>
                    {configItem.value === 'true' ? 'Disable' : 'Enable'}
                  </button>
```

src/pages/index.tsx
```diff
@@ -62,6 +61,16 @@ const AdminPage: NextComponentType = () => {
...
 
+  const handleChange = () => async (): Promise<void> => {
+    await fetch('/api/config', {
+      method: 'POST',
+      body: JSON.stringify({}),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+  };
```
johnkahn, line 72: This endpoint requires that `configKey` and `configValue` are in the body. So you'll need to modify this `handleChange` function to take a parameter and pass those in.
```suggestion
  const handleChange = (configItem: Config) => async (): Promise<void> => {
    const newValue = configItem.value === 'true' ? 'false' : 'true';

    await fetch('/api/config', {
      method: 'POST',
      body: JSON.stringify({
        configKey: configItem.key,
        configValue: newValue
      }),
      headers: {
        'Content-Type': 'application/json',
      },
    });
  };
```

johnkahn (changes requested):
Currently the POST call you are making isn't sending the correct values, that call will fail without them. Also can you put a screenshot of what it looks like?

oscargoddard:
Screen Shot 2020-12-01 at 14 16 40 PM

PR #230, opened 2020-12-01
UI component for editing config items

### LINKED ISSUE #204: [Feature Request] - UI Component for Editing Config Items

oscargoddard:

## Pre-Requisites

## Description of Changes

added the button
added the function to handletoogle

## Related Issues

closes #204

### DIFF
--- a/src/api/config.ts
+++ b/src/api/config.ts
@@ -6,7 +6,7 @@ export const config = express.Router();
 
 config.get('/', async (req, res) => {
   try {
-    res.send(await Config.find());
+    res.send(await Config.find({ order: { key: 'ASC' } }));
   } catch (err) {
     /* istanbul ignore next */
     res.status(500).send('Something went wrong sending an update to users; check the logs for more details');
--- a/src/entities/config.ts
+++ b/src/entities/config.ts
@@ -25,21 +25,4 @@ export class Config extends BaseEntity {
     }
     return toggle ? toggle.value === 'true' : false;
   }
-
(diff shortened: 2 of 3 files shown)


oscargoddard:
My computer died and while I was trying to fix it all my branches are gone. Created another branch and another PR for this issue.

johnkahn (dismissed):

SpencerKaiser (commented):
I don't see a refresh or anything in here... do we need to add that?

johnkahn:
This part handles that
https://github.com/AmericanAirlines/Hangar/pull/230/files#diff-18e0d4553c97cfc420e938ccafb4e3a688e782fa4512ba4ceae3e2a6f24c1987R84-R89

SpencerKaiser:
@johnkahn but what if I have the page open and you change a value? It won't change for me unless I refresh the page and changing the value will "work" but the expected timing will be wrong.

johnkahn:
Ah, I see what you mean

johnkahn (approved):

PR #233, opened 2020-12-04
Common Interface

### LINKED ISSUE #214: [Feature Request] - Common Interface API
### Pre-requisites:

## Describe the problem
We want to be able to add new platforms to the bot without having to rewrite the logic layer for it

A common module API that defines abstracted methods that translate into the platform specific logic. Adding a new platform should be as easy as extending a class and overriding the methods that the platform supports.

Writing code for each platform in the logic with a bunch of `if/else` statements 🤢 

## Functionality to Implement
- A helper method to determine which platform is active. Example pseudocode:
``` typescript
// src/common.ts (or whatever you choose)
// Determine which env vars are present, throw an error if NEITHER or BOTH are supplied
export enum SupportedPlatform {
slack: 'Slack';
discord: 'Discord'
}

export const activePlatform = isSlackActive ? SupportedPlatform.slack : SupportedPlatform.discord;
```
- A single `messageUsers` method (`/src/api/utils/messageUser.ts`) which utilizes the `activePlatform` to figure out which platform util methods should be used

Once implemented, the API should be updated with the new `messageUsers` method. The startup of the app should only try to instantiate a single app (Discord OR Slack) and error should be thrown if NEITHER or BOTH are supplied.

oscargoddard:

## Pre-Requisites

## Description of Changes

Added an environment variable to determine which platform is being used. 
If the PLATFORM_USED is set to "Slack",then only Slack will work. 
If it is set to "Discord", then only Discord will work.

Also, edited the ReadMe.md for the Discord to make people understand the Discord setup easier. 

## Related Issues

Closes #214


johnkahn:
This looks like it's accomplishing the first part of that issue (the helper method to determine which is enabled).
Want to confirm that you can delete the Slack environment variables completely and it'll still start the app with Discord, and vice versa? Since an end user will only have the values for one of them, relevant text from issue below.

> The startup of the app should only try to instantiate a single app (Discord OR Slack) and error should be thrown if NEITHER or BOTH are supplied.

After that's confirmed looks like the last thing is the second part of the issue.

> A single `messageUsers` function which utilizes the `activePlatform` to figure out which platform util methods should be used to send a message
>
> Once implemented, the API should be updated with the new `messageUsers` function.

This will replace the parts of the API that are calling Slack directly to send messages and will instead use this common method so that it works for Discord too.

oscargoddard:
> This looks like it's accomplishing the first part of that issue (the helper method to determine which is enabled).
> Want to confirm that you can delete the Slack environment variables completely and it'll still start the app with Discord, and vice versa? Since an end user will only have the values for one of them, relevant text from issue below.
> 
> > The startup of the app should only try to instantiate a single app (Discord OR Slack) and error should be thrown if NEITHER or BOTH are supplied.
> 
As long as PLATFORM_USED matches the .env vars provided in the .env file it checks out. 

> After that's confirmed looks like the last thing is the second part of the issue.
> 
> > A single `messageUsers` function which utilizes the `activePlatform` to figure out which platform util methods should be used to send a message
> > Once implemented, the API should be updated with the new `messageUsers` function.
> 
> This will replace the parts of the API that are calling Slack directly to send messages and will instead use this common method so that it works for Discord too.

We need a little more help with this. We couldn't really understand how we are going to implement it. We did lots of research though.

johnkahn:
> As long as PLATFORM_USED matches the .env vars provided in the .env file it checks out.

Ah gotcha, perfect 👍 Just wanted to make sure they could leave them blank or something, and that works.

> We need a little more help with this. We couldn't really understand how we are going to implement it. We did lots of research though

So what you're going to want to do is export a function from that file that will check which platform is being used and then either call the Slack API to send a message or the Discord API. It'll look something like this. Haven't tested this, so you'll need to make sure it works. You should be able to register for support through Discord and then accept that request in the admin dashboard. The user should get a DM from the bot saying to come up.

```typescript
export async function sendMessage(userIds: string[], message: string) {
  if (isSlackActive) {
    await messageUsers(userIds, message); // This is imported from src/slack/utilities/messageUsers.ts
  } else {
    await Promise.all(userIds.map(async (id) => {
      const user = await client.users.fetch(id).catch(() => null); // client is imported from src/discord/index.ts

      if (user) {
        user.send(message);
      }
    }));
  }
};
```

TheDivexz:
We were able to get the sendMessage function to not throw any errors. However, this is causing a problem that messages aren't being sent to the user such as when the user needs to be notified when they are up next in a support queue and we have not been able to figure out why that's happening yet. 
image

SpencerKaiser (approved):

PR #239, opened 2021-01-15
Refactor and increasing test coverage

SpencerKaiser:
## Pre-Requisites

## Description of Changes

Refactoring several components within the app, adding test coverage, requiring test coverage for most of the project.


jest.config.js
```diff
@@ -12,16 +12,22 @@ module.exports = {
...
+    '!./src/migration/**',
+    '!./src/tests/**',
+    '!./src/index.ts',
+    '!./src/slack/**',
+    '!./src/discord/index.ts',
+    '!./src/logger.ts',
+    '!./src/app.ts',
+    '!./src/StringDictionary.ts',
+    '!./src/api/admin.ts',
+    '!./src/api/index.ts',
```
johnkahn, line 26: Is the eventual goal to get this down to as few files ignored as possible?

src/api/supportRequest.ts
```diff
@@ -31,18 +31,24 @@ supportRequestRoutes.get('/getAll', async (req, res) => {
...
+    const technicalCount = await SupportRequest.count({ type: SupportRequestType.TechnicalSupport, status: SupportRequestStatus.Pending });
 
-  res.json({
-    ideaCount,
-    technicalCount,
-  });
+    res.json({
+      ideaCount,
+      technicalCount,
+    });
```
johnkahn, line 47: What could fail here?

src/entities/supportRequest.ts
```diff
@@ -72,7 +72,7 @@ export class SupportRequest extends BaseEntity {
     });
 
     const minutesUntilStale = 15;
-
+    // console.log(existingActiveRequests);
```
johnkahn, line 75:
```suggestion
```

johnkahn (changes requested):

src/api/supportRequest.ts
johnkahn: Forgot to include the lines above and below, but you added a `try/catch` around it, but I don't see anything that could fail 🤔

jest.config.js
SpencerKaiser: Some of them, yes. I don't see value in testing some of the files like the logger.

src/api/supportRequest.ts
SpencerKaiser: Potentially these calls since they're hitting the DB: 
``` typescript
const ideaCount = await SupportRequest.count({ type: SupportRequestType.IdeaPitch, status: SupportRequestStatus.Pending });
     const technicalCount = await SupportRequest.count({ type: SupportRequestType.TechnicalSupport, status: SupportRequestStatus.Pending });
```

johnkahn: Looking at their code it looks like it's set to return `0` in a failed call, but I guess a network connection could still fail. In that case though I would think it would have failed before this point. But better safe than sorry haha

johnkahn (approved):

PR #280, opened 2021-02-22
Env fixes

### LINKED ISSUE #258: [Feature Request] - Remove Slack/Discord Specification
### Pre-requisites:

Currently, the user must specify whether they are using Slack or Discord in an environment variable, which could be straightforward.

Upon starting, Hangar should be able to detect which environment variables are set in order to choose between Slack and Discord. For example, if only the `DISCORD_BOT_TOKEN` is set, then Hangar should be spun up using Discord. 

If environment variables for both platforms are set, then an error should be thrown.

### LINKED ISSUE #261: [Feature Request] - Add simple-env
### Pre-requisites:

Currently, environment variables are kind of a pain to deal with. Typos as well as knowing which environment variables are necessary ultimately make it a struggle to set up a .env file. Luckily we can add `simple-env` to trivialize these problems.

We want to set up `simple-env` in Hangar. Details on how to do so can be found [here](https://github.com/AmericanAirlines/simple-env)

### LINKED ISSUE #268: [Bug] - Slack env var is always required
### Pre-requisites:

## Description

Even after specifying `discord` as the chosen platform, the `SLACK_BOT_TOKEN` is still required and the app will crash if it's not provided.

## Expected Behavior

All Slack env vars should be ignored if the chosen platform is Discord

## Actual Behavior

The app fails to start without `SLACK_BOT_TOKEN`

## Steps to Reproduce

Remove `SLACK_BOT_TOKEN`, specify `discord` as the platform, start the app

## Other Information

Workaround: add `SLACK_BOT_TOKEN` to your `.env` and provide `"junk"` as the value:
`SLACK_BOT_TOKEN="junk"`

Ulfingrad:
## Pre-Requisites

## Description of Changes

Implements simple-env along with a couple of other minor fixes such as removing the need for a Slack token in addition to removing the need for platform specification.

## Related Issues

Closes #268
Closes #261 
Closes #258


src/tests/api/config.test.ts
```diff
@@ -6,6 +6,17 @@ import { createDbConnection, closeDbConnection } from '../testdb';
 const adminSecret = 'Secrets are secretive';
```
johnkahn, line 6: We should delete this. If it's being used we should import `env` above and then access `env.adminSecret`

johnkahn: Looks like this pattern is in some other files as well, we should make sure that tests are using the imported `env`

johnkahn (changes requested):

johnkahn (approved):

PR #318, opened 2021-03-16
replaced admin/support/judge secret access from .env with the database

RazvanPreotu99:
## Pre-Requisites

## Description of Changes

Gets admin/support/judge secrets from database instead of .env file.
Does this look to be on the right track? We are having problems fixing the tests as the line
const adminSecret = await Config.findOneOrFail("adminSecret"); 
in requireAuth.ts breaks them.  

Also when importing {Config} in support/[secret].tsx and judge/[secret].tsx we get an error when visiting those pages. A picture showing the error is 
support and judge config error
attached.

## Related Issues

Closes #318  - Access secrets from config table in database instead of .env

Co-authored by: <Tatum Hutton 78566681+tatumwashere@users.noreply.github.com>


SpencerKaiser:
@RazvanPreotu99 please make sure to update your description to use the `Closes` syntax so your issue becomes linked to your PR

SpencerKaiser:
@tatumwashere remember to be the committer for one of the two stories on the next iteration

src/api/admin.ts
```diff
@@ -1,12 +1,17 @@
...
 
 export const admin = express.Router();
 
 admin.use(express.json());
 
-admin.post('/login', (req, res): void => {
-  if (req.body.secret !== env.adminSecret) {
+admin.post('/login', async (req, res): Promise <void> => {
+  const adminSecret = await Config.findOneOrFail('adminSecret');
+  if(adminSecret.value === ''){
```
SpencerKaiser, line 10: This will include empty strings, undefined, and null
```suggestion
  if(!adminSecret.value){
```

src/api/admin.ts
```diff
@@ -1,12 +1,17 @@
...
+import { Config } from '../entities/config';
 
 export const admin = express.Router();
 
 admin.use(express.json());
 
-admin.post('/login', (req, res): void => {
-  if (req.body.secret !== env.adminSecret) {
+admin.post('/login', async (req, res): Promise <void> => {
+  const adminSecret = await Config.findOneOrFail('adminSecret');
```
SpencerKaiser, line 9: Without a try/catch here, the app would become unusable in this state. For now, let's just use `findOne` and we can deal with the other half of this flow later

src/api/admin.ts
```diff
@@ -1,12 +1,17 @@
...
+  if(adminSecret.value === ''){
+    res.status(401).send('Secret not initialized');
```
SpencerKaiser, line 11: This is a fault the server side of things, so it should be a 500
```suggestion
    res.status(500).send('Secret not initialized');
```

Also please add a `logger.crit` in here to log out what went wrong

src/api/middleware/requireAuth.ts
```diff
@@ -1,9 +1,12 @@
 import { Handler } from 'express';
 import logger from '../../logger';
-import { env } from '../../env';
+import { Config } from '../../entities/config';
 
-export const requireAuth = (redirect = false): Handler => (req, res, next): void => {
-  if (req.signedCookies?.authed === 'yes' || req.headers.authorization === env.adminSecret || req.headers.authorization === env.supportSecret) {
+export const requireAuth = (redirect = false): Handler => async (req, res, next): Promise<void> => {
+  const adminSecret = await Config.findOneOrFail("adminSecret");
+  const supportSecret = await Config.findOneOrFail("supportSecret");
```
SpencerKaiser, line 7: Use `findOne` here as well

SpencerKaiser (changes requested):
John is going to review the react stuff shortly 😉

src/pages/judge/[secret].tsx
```diff
@@ -1,236 +1,239 @@
...
-  }
-
-  return {};
-};
-
-export default judge;
+/* globals localStorage, fetch */
+
+import React from 'react';
+import { Config } from '../../entities/config';
```
Comment (johnkahn) [inline src/pages/judge/[secret].tsx:4]: Was this causing any errors in your terminal or in the page's console when the page is loaded? I feel like it might, but if it doesn't then that's fine.

src/pages/judge/[secret].tsx
```diff
@@ -1,236 +1,239 @@
...
+        </>
+      )}
+    </div>
+  );
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+judge.getInitialProps = async (ctx: any): Promise<any> => {
+  const judgeSecret = await Config.findOneOrFail('judgeSecret');
+  if (ctx.res && ctx.query.secret !== judgeSecret.value) {
```
Comment (johnkahn) [inline src/pages/judge/[secret].tsx:229]: `ctx.res` will only be there on the server, so we're basically saying if we're on the server and the secret does not match, throw a 404. Because of that we only want to reach out to the database if we're on the server.

```suggestion
  if (ctx.res && ctx.query.secret !== judgeSecret.value) {
  const judgeSecret = await Config.findOneOrFail('judgeSecret');
```

src/pages/support/[secret].tsx
```diff
@@ -91,13 +91,14 @@ const support: NextPage<{ secret: string }> = (props): JSX.Element => {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 support.getInitialProps = async (ctx: any): Promise<any> => {
-  if (ctx.res && ctx.query.secret !== env.supportSecret) {
+  const supportSecret = await Config.findOneOrFail('supportSecret');
+  if (ctx.res && ctx.query.secret !== supportSecret.value) {
```
Comment (johnkahn) [inline src/pages/support/[secret].tsx:95]: Same as the other

```suggestion
  if (ctx.res && ctx.query.secret !== supportSecret.value) {
  const supportSecret = await Config.findOneOrFail('supportSecret');
```

johnkahn (changes requested):

src/pages/judge/[secret].tsx
Comment (tatumwashere) [reply src/pages/judge/[secret].tsx:229]: We're confused on this suggestion because if we switch the two lines then we get an error since judgeSecret.value is not yet defined. Also, if judgeSecret.value is defined only within the if statement then it messes up the return statement as well since it is not defined outside of the if statement.

RazvanPreotu99:
We get testing errors from team, supportRequest and health but I don't think it's from our changes?

src/pages/support/[secret].tsx
```diff
@@ -91,13 +90,21 @@ const support: NextPage<{ secret: string }> = (props): JSX.Element => {
...
+    // if (ctx.query.secret !== judgeSecret.value) {
+
+    //   ctx.res.statusCode = 404;
+    //   ctx.res.end('Not found');
+    // }
 
+    // return {
+    //   secret: judgeSecret.value
+    // };
+  }
```
Comment (SpencerKaiser) [inline src/pages/support/[secret].tsx:105]: @johnkahn can you provide feedback on what to do here?

Comment (johnkahn) [reply src/pages/support/[secret].tsx:105]: Remove them

SpencerKaiser (changes requested):
Quick fixes

src/pages/support/[secret].tsx
Comment (johnkahn) [reply src/pages/support/[secret].tsx:105]: The if statement too, just the whole thing should be deleted

PR #321, opened 2021-03-17
Added buttons to the support table

### LINKED ISSUE #313: [Feature Request] - Add Buttons to Support Table (Help, Complete, No-Show, Remind)
### Pre-requisites:

Related to #312, add buttons to the table.

Add buttons to the end of each row so a user can switch the status of a request. Request options are help, complete, no-show, and remind.

OlinZ:
## Pre-Requisites

## Description of Changes

Added 4 buttons (select, no show, complete, remind) to each row in the table with working functionalities. The next steps are to add logic so only some of the buttons show up at a given time.

## Related Issues

Related to #312 
Closes #313


OlinZ:
Here is what the table currently looks like:
image

Made changes to the `supportRequest/getSpecific` API to update `movedToInProgressAt`, `syncHash`, and `supportName` in the database when assigning requests.

src/tests/api/supportRequest.test.ts
```diff
@@ -413,20 +413,26 @@ describe('api/supportRequest', () => {
...
-      expect(mockQueryBuilder.set).toBeCalledWith({ status: SupportRequestStatus.InProgress });
-      expect(mockQueryBuilder.where).toBeCalledWith({ id: supportRequest.id, status: SupportRequestStatus.Pending });
-      expect(sendMessageSpy.mock.calls[0][0]).toEqual([supportRequest.slackId]);
+      // expect(mockQueryBuilder.set).toBeCalledWith({
+      //   status: SupportRequestStatus.InProgress,
+      //   supportName: supportRequest.supportName,
+      //   movedToInProgressAt: supportRequest.movedToInProgressAt,
+      // });
+      // expect(mockQueryBuilder.where).toBeCalledWith({ id: supportRequest.id, status: SupportRequestStatus.Pending });
+      // expect(sendMessageSpy.mock.calls[0][0]).toEqual([supportRequest.slackId]);
```
johnkahn, line 435: Are these no longer needed? If not we can just delete them, but I'm pretty sure we would still want to make sure that these things are called correctly.

src/api/supportRequest.ts
```diff
@@ -186,10 +187,14 @@ supportRequestRoutes.patch('/getSpecific', async (req, res) => {
...
   }
   try {
+    const newHash = genHash();
     await SupportRequest.createQueryBuilder('supportRequest')
       .update()
       .set({
         status: SupportRequestStatus.InProgress,
+        movedToInProgressAt: new Date(),
+        syncHash: newHash,
+        supportName,
```
johnkahn, line 197: Why were these added? 🤔 I may have forgotten if we talked about it, but I'm curious why they weren't there before.

johnkahn (changes requested):
LGTM, just a couple questions 👍

src/api/supportRequest.ts
OlinZ: `/supportRequest/getSpecific` was not updating those fields in the database and was causing errors when I tried to assign items to people. My guess is since `/getSpecific` was never really used before (old code only used `/getNext`), this part of the code was never tested.

src/tests/api/supportRequest.test.ts
OlinZ: Yes these are still needed. It was giving an error at `movedToInProgressAt: supportRequest.movedToInProgressAt` since `new Date()` returns the current time and there's always a delay between the actual API call and the jest check. Not sure how to proceed but I think we can try mocking the `new Date()` function.

OlinZ: Update: I spent some time on this and was able to mock both `new Date()` and `genHash()`. Should be good now.

PR #335, opened 2021-03-30
Fixing: Eslint is not checking tsx files -  #217

### LINKED ISSUE #217: [Bug] - Eslint is not checking tsx files
### Pre-requisites:

## Description

Our implementation of prettier is ignoring tsx files

## Expected Behavior

All relevant code should be linted

## Actual Behavior

The files are not being linted

## Steps to Reproduce

Make a change to a tsx file that should throw an eslint warning/error and then look at the code within VSCode (with the eslint plugin enabled)

dre8597:
## Pre-Requisites

## Description of Changes
Added "tsx" & "jsx" to the eslint accepted extensions
Added "jsx: preserved" to allow tsx translation to jsx without error 
Fixed typing error in nodemonConfig delay

## Related Issues
Closes #217

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build:ts": "tsc -p src/ --pretty",
     "build:next": "next build",
     "build:watch": "npm run build:ts -- -w",
-    "lint": "eslint \"src/**/*.ts\" --max-warnings 0",
+    "lint": "eslint --ext={js,jsx,ts,tsx} src --max-warnings 0",
     "lint:fix": "npm run lint -- --fix",
     "lint:staged": "lint-staged",
     "checkstyle": "prettier -l \"src/**/*.ts\"",
@@ -57,7 +57,7 @@
     ]
   },
   "nodemonConfig": {
-    "delay": "400"
+    "delay": 400
   },
   "dependencies": {
     "@americanairlines/simple-env": "^1.0.2",
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -11,7 +11,8 @@
     "forceConsistentCasingInFileNames": true,
     "lib": ["esnext", "esnext.asynciterable"],
     "emitDecoratorMetadata": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "jsx": "preserve",
   },
   "include": ["./**/*.ts"],
   "exclude": ["node_modules", "./**/*.test.ts"]
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,7 +28,7 @@ module.exports = {
   settings: {
     'import/resolver': {
       node: {
-        extensions: ['.js', '.ts'],
+        extensions: ['.js', '.ts', '.tsx', '.jsx'],
       },
     },
     react: {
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "build:ts": "tsc -p src/ --pretty",
     "build:next": "next build",
     "build:watch": "npm run build:ts -- -w",
-    "lint": "eslint \"src/**/*.ts\" --max-warnings 0",
+    "lint": "eslint --ext={js,jsx,ts,tsx} src --max-warnings 0",
```
johnkahn, line 14: Out of curiosity do we need to have the `ext` flag if the extensions are also listed on line 31 of the `.eslintrc.js` file? If we can remove the flag that would be great, then we have them in one place. If it doesn't work by removing the flag though that's fine.

johnkahn (changes requested):
LGTM, just a quick question about the flag

Also looks like that flag needs to be a string anyway 🤔 If we can remove it and it works that would probably just be best to keep it in one place.

package.json
dre8597: Ok, I fixed the flag to be a string. From my understanding, the `.eslintrc.js` is what the ide uses to know the eslint rules, and anything in `package.json` is used for building and compiling but I may be wrong.

johnkahn: Sounds good 👍

johnkahn (changes requested):
One last thing before we can merge, can you add yourself to the `Authors.md` file?

johnkahn (changes requested):
Let's leave the eslint config in the separate file, we have nodemon in there because it's pretty small, but generally we keep configuration in their own files so the package.json can be just the info, dependencies, and scripts.

> Husky and lint-staged are exceptions that had to be in there, but those will actually be removed for different reasons soon.

dre8597:
I found that the eslint configs can be placed in an eslintrc.js file or package.json so I placed them all into package.json for consolidation.

johnkahn (dismissed):
LGTM 👍

src/tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "forceConsistentCasingInFileNames": true,
     "lib": ["esnext", "esnext.asynciterable"],
     "emitDecoratorMetadata": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "jsx": "preserve",
```
SpencerKaiser, line 15:
```suggestion
    "experimentalDecorators": true
```

tsconfig.json
```diff
@@ -9,9 +9,8 @@
     "strict": false,
     "noEmit": true,
     "resolveJsonModule": true,
-    "isolatedModules": true,
-    "jsx": "preserve"
-  },
+    "isolatedModules": true
```
SpencerKaiser, line 12:
```suggestion
    "isolatedModules": true,
    "jsx": "preserve"
```

SpencerKaiser (changes requested):

tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve"
-  },
+    },
```
SpencerKaiser, line 14:
```suggestion
  },
```

Issue #231, opened 2020-12-03
[Feature Request] - Remove SQLite in tests


johnkahn:
### Pre-requisites:
- [x] I looked through both [open and closed issues](../issues?utf8=✓&q=is%3Aissue) and did not find another request for the same or similar feature.

## Is your feature request related to a problem, or an open issue? Describe the problem, and if applicable, reference the issue.
We have to bend over backwards to get our production code working in the test environment with SQLite.

## Describe the solution you'd like
* Remove SQLite from tests.
* Create a mock `BaseEntity` that all the mocked `Entity` classes can use. This will setup the common mocks for `save`, `findOne`, etc. Then create a mock file for each that correctly mock out static/instance methods that are specific to that class. Then call `jest.mock()` on each entity in the setup 
* Change all instances of `simple-json` to `json`. (Along with any enums, arrays, and etc SQLite can't do natively)

## Describe alternatives you've considered
S̸͙̓Ṷ̵̗̝̼͒͆͌F̸̱͉͙͌̒F̵̫̀̑͋Ȩ̷̭̓̊͒̎R̸͈̾͠

## Additional context

<img src="https://user-images.githubusercontent.com/7807353/101080625-a2dd0300-356e-11eb-8ab3-713180eab598.png" width="300" />

SpencerKaiser:
Already fixed by #319

PR #337, opened 2021-03-31
Updated Tests for Team to Mock Database.

OlinZ:
## Pre-Requisites

## Description of Changes

Updated `team.test.ts` to mock database functions. Need some feedback since we are not sure if we went in the right direction. 

## Related Issues

Related to #231 

`Co-authored-by: Evelio Ramos <44717805+eveliososa@users.noreply.github.com>`

### DIFF
--- a/src/entities/team.ts
+++ b/src/entities/team.ts
@@ -54,7 +54,6 @@ export class Team extends BaseEntity {
 
     /* eslint-disable no-await-in-loop */
     do {
-      console.log('Retries: ', retries);
       team = await Team.findOne({
         where: {
           id: Not(In(excludedTeamIds)),
@@ -68,10 +67,9 @@ export class Team extends BaseEntity {
       if (team) {
         const newHash = genHash();
         const result = await Team.updateSelectedTeam(team, newHash);
-        console.log('IN METHOD:', result);
         await team.reload();
 
-        if (result.affected > 0 || team.syncHash === newHash) {
+        if (result.affected > 0) {
           // We found a team and assigned the judge correctly; return it
           return team;
         }
@@ -90,7 +88,7 @@ export class Team extends BaseEntity {
     throw new Error('Unable to retrieve a team due to concurrency issues');
   }
 
-  // TODO: Move to a util method outside of this file
+  /* istanbul ignore next */
   static async updateSelectedTeam(team: Team, hash: string): Promise<UpdateResult> {
     return Team.createQueryBuilder()
       .update()
@@ -100,6 +98,7 @@ export class Team extends BaseEntity {
       .execute();
   }
 
+  /* istanbul ignore next */
   async decrementActiveJudgeCount(): Promise<void> {
     await Team.createQueryBuilder('team')
       .update()
@@ -108,6 +107,7 @@ export class Team extends BaseEntity {
       .execute();
   }
 
+  /* istanbul ignore next */
   async incrementJudgeVisits(): Promise<void> {
     await Team.createQueryBuilder('team')
       .update()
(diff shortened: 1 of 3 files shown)


src/tests/entities/team.test.ts
```diff
@@ -1,63 +1,147 @@
...
+const mockQueryBuilder = {
+  where: jest.fn().mockReturnThis(),
+  clone: jest.fn().mockReturnThis(),
+  andWhere: jest.fn().mockReturnThis(),
+  orderBy: jest.fn().mockReturnThis(),
+  addOrderBy: jest.fn().mockReturnThis(),
+  getOne: jest.fn().mockReturnThis(),
+};
+
+describe('', () => {
```
SpencerKaiser, line 40: We need to pass a string in here to describe what we're testing. Looks like the previous one was `judging`... should we add that back here?

src/tests/entities/team.test.ts
```diff
@@ -1,63 +1,147 @@
...
+    team.activeJudgeCount = 2;
+    mockCreateQueryBuilder.execute = jest.fn(() => {
+      team.activeJudgeCount = 1;
+    });
     await team.decrementActiveJudgeCount();
-    const updatedTeam = await Team.findOneOrFail(team.id);
-    expect(updatedTeam.activeJudgeCount).toEqual(0);
+    expect(createQuerySpy).toBeCalledTimes(1);
+    expect(createQuerySpy).toBeCalledWith('team');
+    expect(team.activeJudgeCount).toEqual(1);
```
SpencerKaiser, line 56: This really isn't accomplishing much... your `expect`s here are mostly just checking that the query builder was created for the right entity and that your mock method returns the value you asked it to return. Instead, we should be checking other things like `Was the set method called with an object that includes a method with that appropriate name that attempts to decrement the count?`.

src/tests/entities/team.test.ts
```diff
@@ -1,63 +1,147 @@
...
     await team.incrementJudgeVisits();
-    let updatedTeam = await Team.findOneOrFail(team.id);
-    expect(updatedTeam.judgeVisits).toEqual(1);
-
-    await team.incrementJudgeVisits();
-    updatedTeam = await Team.findOneOrFail(team.id);
-    expect(updatedTeam.judgeVisits).toEqual(2);
+    expect(createQuerySpy).toBeCalledTimes(1);
+    expect(createQuerySpy).toBeCalledWith('team');
+    expect(team.judgeVisits).toEqual(2);
```
SpencerKaiser, line 67: Same comment as the test above;

SpencerKaiser (changes requested):
I think many of the tests aren't testing the functionality and are instead testing the test itself... if you want to see if your test was written well, change the implementation significantly and see if the test break. For example, if you modify this line:
``` typescript
      .set({ activeJudgeCount: () => '"team"."activeJudgeCount" - 1' })
```

and change the decrement to 12 (`'"team"."activeJudgeCount" - 1'`), your tests would still pass...

Take another pass, fix the lint errors, and let us know when you're ready for a new review 🙂

src/entities/team.ts
```diff
@@ -95,6 +88,7 @@ export class Team extends BaseEntity {
     throw new Error('Unable to retrieve a team due to concurrency issues');
   }
 
+  /* istanbul ignore next */
   static async updateSelectedTeam(team: Team, hash: string): Promise<UpdateResult> {
```
SpencerKaiser, line 92: @johnkahn I realized that 99% of these methods are just sql... I guess we could check for the values the use, but I'm not sure I see much value in that. Thoughts?

johnkahn (approved):

Issue #260, opened 2021-02-03
[Feature Request] - Create functionality for DB seeding


Ulfingrad:
### Pre-requisites:
- [x] I looked through both [open and closed issues](../issues?utf8=✓&q=is%3Aissue) and did not find another request for the same or similar feature.

## Is your feature request related to a problem, or an open issue? Describe the problem, and if applicable, reference the issue.

We are currently manually adding config items whenever we need them (which isn't great long-term), so we need an easier way for them to be added to the db.

## Describe the solution you'd like

Ultimately, we should have a migration that runs through a  json containing a list of all the config items that we have and adds them to the db. To complement this, we should have a new npm script (we can call it `seedconfig` or something) that runs the migration when executed.

Notes:
- Check out [`typeorm seeding`](https://github.com/w3tecch/typeorm-seeding) as a good library for making this happen
- Make sure to update the README to include instructions

PR #341, opened 2021-04-05
Seed Database w/ Dummy Team and Judge Data

OlinZ:
## Pre-Requisites

## Description of Changes

Used [typeorm-seeding](https://github.com/w3tecch/typeorm-seeding) to seed db with fake team and judge data. We had to change `"module": "esnext"` to `"module": "commonjs"` so `typeorm-seeding` compiles and runs (not sure if this is the right solution). 

To seed the db: run `npm run seed:run` and it will generate 10 fake teams and 10 fake judges.

## Related Issues

Related to #260 

`Co-authored-by: Evelio Ramos <44717805+eveliososa@users.noreply.github.com>`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4189,6 +4189,11 @@
         "@types/range-parser": "*"
       }
     },
+    "@types/faker": {
+      "version": "5.5.0",
+      "resolved": "https://registry.npmjs.org/@types/faker/-/faker-5.5.0.tgz",
+      "integrity": "sha512-WPBf6jgCsRrbPrgDuWHfbI+cd2CT33JUK+w8NM8jU7KmNr9PF+2eKhD0DKV32nMlyzdMnTHc5TSAG1jpkZbN5A=="
+    },
     "@types/graceful-fs": {
       "version": "4.1.5",
       "resolved": "https://registry.npmjs.org/@types/graceful-fs/-/graceful-fs-4.1.5.tgz",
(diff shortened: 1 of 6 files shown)


johnkahn:
Looks good, what was the error that you were getting before you changed the module?

OlinZ:
> Looks good, what was the error that you were getting before you changed the module?

Something about `Cannot use import statement outside a module`. Here's a screenshot of the error.

image

johnkahn:
Gotcha, so it's because those files don't export anything. Could you put the `define` code in the entity itself? Like at the bottom of the file, I think you'll need to modify the configuration in the ormconfig file to point the seed factories to the entities instead of the factories. But I think that'll work and let us keep `esnext` as our module type.

OlinZ:
I tried that and it still throws the same error.
Here's what I did:
- Moved the contents of `judge.factory.ts` and `team.factory.ts` to below the code in `src/entities/team.ts` and `src/entities/judge.ts`.
- Added `"factories": ["src/entities/team.ts", "src/entities/judge.ts"]` to `ormconfig.json`.
- Changed `"module": "commonjs"` back to "module": "esnext"

Here's the error:
image

src/seed.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import { createConnection, getConnectionOptions, getConnection, ConnectionOptions } from 'typeorm';
+import { PostgresConnectionOptions } from 'typeorm/driver/postgres/PostgresConnectionOptions';
+import path from 'path';
+import { Team } from './entities/team';
+import { Judge } from './entities/judge';
+import { env } from './env';
+import logger from './logger';
+
+async function initDatabase(): Promise<void> {
```
johnkahn, line 11: You should be able to just import and call the same initDatabase that the app does so you don't need to write your own

johnkahn (changes requested):
I think it's good to go, just a small change, and it's hard to review because of all the code cov warnings, can you add `!src/seed.ts` to the coverage ignore section of `jest.config.js` file? That will make it easier to review.

image

src/seed.ts
OlinZ: I think don't the code exports `initDatabase` anywhere. I just copied this function from `app.ts`. I guess we can export `initDatabse` in `app.ts` and import it here, let me know what you think.

johnkahn: Gotcha, yeah let's export that and use it here 👍

PR #343, opened 2021-04-06
Config refactor

### LINKED ISSUE #299: [Feature Request] - Modify config endpoint to return all config items AND default items
### Pre-requisites:

This implementation should start by creating a new enum (potentially in a new file called `src/types/config.ts`) which tracks `KnownConfig` items like `adminSecret` and `judgeSecret`.

The `/config` api endpoint should be modified to return the full set of known config items (saved in the `Config` table) AND the list of default items (`KnownConfig`). If a default key is already set, the default value should be overridden by what's in the DB.

For example, if the `Config` table has `adminSecret` set to `true` but nothing else, the response should include all the OTHER config items as well (pseudocode)
``` typescript
return {
  adminSecret: true,       // Came from Config Table
  judgeSecret: undefined,  // Was not found in Config table but is a `KnownConfig` item
  // ... all other `KnownConfig` items with a value of `undefined`
}
```

## Current Solution
`GET` `localhost:3000/api/config`: `[{"key":"supportRequestQueueActive","value":"true"},{"key":"teamRegistrationActive","value":"true"}]`

## Desired Solution
`GET` `localhost:3000/api/config`: `[{"key":"supportRequestQueueActive","value":"true"},{"key":"teamRegistrationActive","value":"true"}, {"key":"adminSecret","value":"undefined"}, {"key":"SOME_OTHER_KNOWN_CONFIG","value":"true"}, ... ]`
Where `adminSecret` is not already present in the `Config` table

dre8597:
## Pre-Requisites
- [X] Yes, I included and/or modified tests to cover relevant code **OR** my change is non-technical

## Description of Changes

Adding logic to always display list of known config items even when they are not set

## Related Issues

Closes #299


src/api/config.ts
```diff
@@ -1,12 +1,19 @@
...
 import logger from '../logger';
+import { KnownConfig } from '../types/config';
 
 export const config = express.Router();
 
 config.get('/', async (req, res) => {
   try {
-    res.send(await Config.find({ order: { key: 'ASC' } }));
+    const configItems: Config[] = await Config.find();
+    Object.values(KnownConfig).forEach((knownConfigItem): void => {
```
johnkahn, line 11: If we're not using the keys of the enum, let's just make it an array at the top of this file and call it `knownConfigKeys`

src/api/config.ts
```diff
@@ -1,12 +1,19 @@
...
+      if (!configItems.find((configItem: Config) => configItem.key === knownConfigItem)) {
+        configItems.push(new Config(knownConfigItem, 'undefined'));
```
johnkahn, line 13: Will it let you put `null` as the value? I think a change that the UTD students did should allow that and a frontend change I have coming soon will show that properly.

```suggestion
        configItems.push(new Config(knownConfigItem, null));
```

johnkahn (changes requested):
Two quick changes

src/api/config.ts
johnkahn: Never mind, enum has other purposes

johnkahn (approved):

Issue #338, opened 2021-04-01
[Feature Request] - Move remaining .env entries to config table


tristengeven:
### Pre-requisites:
- [x] I looked through both [open and closed issues](../issues?utf8=✓&q=is%3Aissue) and did not find another request for the same or similar feature.

## Describe the solution you'd like

In preparation for #330 the remaining env values should be moved into the ConfigItems database/entity. Areas in app which originally requested from the .env should instead communicate with the database.

PR #348, opened 2021-04-08
Slack configuration values pulled from database

tristengeven:
## Pre-Requisites

## Description of Changes

Change queries for slack env values to database instead of .env

## Related Issues

Partially handles #338


johnkahn:
@tristengeven I think there needs to be more changes to this before we can resolve it 🤔  Looks like there are some compilation errors

tristengeven:
I believe this should be closed in favor of #356

PR #415, opened 2021-11-02
adding check to ensure user is a member of the guild

### LINKED ISSUE #382: Retrieve Guild information on account creation
### Pre-requisites:

## Description

When a user initially logs into the app, we need to make a call to the Discord API to retrieve their guild information to make sure they are authenticated with the appropriate server

If the user is _**not**_ part of the server, throw a 403 and redirect to an appropriate error page

dre8597:
## Pre-Requisites

## Description of Changes

Add check to see if a user is a part of the Discord Guild prior to providing a session to the user.

## Related Issues

Closes #382


packages/api/src/index.ts
```diff
@@ -38,7 +38,13 @@ void (async () => {
           await entityManager.persistAndFlush(newUser);
         }
 
-        done(null, { profile, accessToken, refreshToken });
+        const guildIds = profile.guilds?.filter((guild) => guild.id === env.discordGuildId) ?? [];
+
+        if (guildIds.length > 0) {
+          done(null, { profile, accessToken, refreshToken });
+        } else {
+          done(new Error('Please make sure your are about of the event discord server'));
```
johnkahn, line 46: Does this give a 403 automatically?

```suggestion
          done(new Error('Please make sure your are a member of the Discord server for this event'));
```

dre8597: It does not, I tried using the Discord ```AuthorizationError``` where it allowed me to attach the status/code but that threw an internal error of an uncaught exception for the error type.

packages/api/src/env.ts
```diff
@@ -15,6 +15,7 @@ export const env = setEnv({
     appUrl: 'APP_URL',
   },
   optional: {
+    discordGuildId: 'DISCORD_GUILD_ID',
```
SpencerKaiser, line 18: Make sure to update the `.env.sample` file

packages/api/src/index.ts
```diff
@@ -38,7 +38,15 @@ void (async () => {
           await entityManager.persistAndFlush(newUser);
         }
 
-        done(null, { profile, accessToken, refreshToken });
+        const guildIds = profile.guilds?.filter((guild) => guild.id === env.discordGuildId) ?? [];
```
SpencerKaiser, line 41: I'd prefer something like this only because `guildIds` implies that it's ALL of the guild IDs and it makes line 43 ambiguous (it currently reads like `if the user is in a guild, they're good to go`.

@johnkahn thoughts? Is there a better way to do the `map` business?

```suggestion
        const isMemberOfEventGuild = (profile.guilds ?? []).map((guild) => guild.id).includes(env.discordGuildId);
```

packages/api/src/index.ts
```diff
@@ -38,7 +38,15 @@ void (async () => {
...
+
+        if (guildIds.length > 0) {
```
SpencerKaiser, line 43: Just for future reference, anything greater than 0 is truth, so you really just need this:

```suggestion
        if (guildIds.length) {
```

SpencerKaiser (changes requested):

packages/api/src/index.ts
johnkahn: Hm, what if you create a new `PassportVerifyError` class that extends `Error`. Then you can pass that to `done`. You would need to set a new error handler though right before `app.listen` is called.

```typescript
export class PassportVerifyError extends Error {
  constructor(message: string) {
    super(message);
  }
}
```

```typescript
// After all other `app.use/get/post/etc` calls, before `app.listen`
app.use((req, res, next, err) => {
  if (err instanceof PassportVerifyError) {
    res.status(403).send(err.message);
    return;
  }

  next();
});
```

packages/api/src/index.ts
johnkahn: Nope, looks good to me

johnkahn (approved):

PR #444, opened 2021-11-13
Queue user post endpoint

### LINKED ISSUE #430: Create QueueUser `POST` endpoint
### Pre-requisites:

## Description

We need a new API endpoint which allows authenticated users to join help queues

- Create post request `/api/queue/join` that will create a new `QueueUser`
- It will take a single body key `type` which will be one of the values of the `QueueType` enum
- The error message for an invalid `type` should be generic like `Invalid queue type` so people can't figure out what queues we have enabled

Ulfingrad:
## Pre-Requisites

## Description of Changes

Adds the endpoint responsible for adding users to a queue.

## Related Issues

Closes #430


packages/api/src/api/queue.ts
```diff
@@ -27,3 +29,39 @@ queue.get('/:type', async (req, res) => {
     res.status(500).send(errorMsg);
   }
 });
+
+queue.post('/join', async (req, res) => {
```
SpencerKaiser, line 33: I'm kinda thinking this makes more sense without the `join`... thoughts?

packages/api/src/api/queue.ts
```diff
@@ -27,3 +29,39 @@ queue.get('/:type', async (req, res) => {
...
+  if (!(type in QueueType)) {
+    res.status(400).send('The queue type entered is invalid');
+    return;
+  }
+
+  try {
+    const currentQueueItems = await req.entityManager.find(QueueUser, {
+      user,
+      status: { $in: ['Pending', 'InProgress'] },
+    });
```
SpencerKaiser, line 46: Let's change this call to a count, that way call is faster
```suggestion
    const userAlreadyInQueue = await req.entityManager.count(QueueUser, {
      user,
      status: { $in: ['Pending', 'InProgress'] },
    }).then((count) => !!count);
```

SpencerKaiser (commented):
Super tiny change and one question

dre8597 (approved):
LGTM 🚀

PR #475, opened 2023-08-01
Sign in with Slack

granthill7:
## Pre-Requisites


SpencerKaiser:
@granthill7 - I had a small callout in the working agreement; please open PRs as a draft until you're ready for a formal review (which will eliminate the need for `IN PROGRESS` in the title and cut down on premature Slack notifications 😉)

packages/api/src/api/auth/slack.ts
```diff
@@ -0,0 +1,15 @@
...
+export const slack = async (req: Request, res: Response) => {
+  const linkBase: string =
+    'https://slack.com/openid/connect/authorize?scope=openid%20email%20profile&response_type=code';
+  // Make this an env variable
+  const redirect: string =
+    '&redirect_uri=https://1376-162-92-127-85.ngrok-free.app/api/callback/slack';
+  // Does this need to be env?
+  const client: string = `&client_id=${env.slackClientID}`;
+
+  const fullLink = linkBase + redirect + client;
```
SpencerKaiser, line 13: Can you move all this up out of the function? There's no reason to keep redeclaring
```suggestion
  const slackAuthBaseUrl: string =
    'https://slack.com/openid/connect/authorize?scope=openid%20email%20profile&response_type=code';
  const queryArgs = new URLSearchParams({
    'redirect_uri': `${env.baseUrl}/api/callback/slack`,
    'client_id': env.slackClientID,
  });

  const fullLink = `${slackAuthBaseUrl}${queryArgs.toString()}`;
```

packages/api/src/api/callback/index.ts
SpencerKaiser, line 1: This route needs `/auth` after `/api`

packages/api/src/api/callback/slack.ts
SpencerKaiser, line 1: Same for this path

packages/api/src/api/callback/slack.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  return { email, firstName, lastName };
+};
+
+export const slack = async (req: Request, res: Response) => {
+  const myCode: string = req.query.code as string;
+  const { slackClientID, slackClientSecret } = env;
+
+  // Make this an env variable
+  const redirect: string = 'https://1376-162-92-127-85.ngrok-free.app/api/callback/slack';
```
SpencerKaiser, line 23:
```suggestion
  const redirect: string = `${env.baseUrl}/api/callback/slack`;
```

packages/api/src/api/index.ts
```diff
@@ -12,6 +14,8 @@ if (env.nodeEnv !== 'development') api.use(enforceRateLimiting);
 
 // UNPROTECTED ROUTES
 api.use('/health', health);
+api.use('/auth', auth);
+api.use('/callback', callback);
```
SpencerKaiser, line 18: This should moved into the auth base route

packages/api/src/env.ts
```diff
@@ -7,8 +7,11 @@ export const env = setEnv({
   required: {
     nodeEnv: 'NODE_ENV',
     port: 'PORT',
+    baseURL: 'BASE_URL',
```
SpencerKaiser, line 10: I _think_ this is my opinion; thoughts?
```suggestion
    baseUrl: 'BASE_URL',
```

packages/api/src/api/index.ts
granthill7: So `/callback` should be moved into `api/auth/index.ts`?

SpencerKaiser: It should be registered in that file so the full url would be `/api/auth/callback/slack`

SpencerKaiser (approved):

PR #476, opened 2023-08-02
Oauth validate

alexanderallen-aa:
## Pre-Requisites


packages/api/src/utils/authenticateUser.ts
```diff
@@ -0,0 +1,25 @@
+import { Response , Request } from 'express';
+
+type OAuthUserData = {
+    email: string ,
+    firstName?: string ,
+    lastName?: string ,
+    phoneNumber?: string ,
+    address?: string ,
+    given_name?: string ,
+    family_name?: string ,
```
SpencerKaiser, line 10: First/last should be necessary for everything and should be accessible via the APIs for each auth provider so let's make those required. Beyond that, I'd prefer to start with just what we need and expand from there
```suggestion
    firstName: string ,
    lastName: string ,
```

packages/api/src/utils/authenticateUser.ts
```diff
@@ -0,0 +1,25 @@
...
+    // if not, redirect to error
+    if (options.email) {
+        req.session = options
+        res.redirect('/')
+    } else {
+        res.redirect('/error')
+    }
+}
+
+export { authenticateUser , OAuthUserData }
```
SpencerKaiser, line 25: While this syntax is totally valid, I highly prefer `export` to be used where the items are being declared as it increases transparency around what is and what is not exported. If you feel strongly about this syntax, let's discuss as a team but I want to standardize one way or the other.
```suggestion
```

packages/api/tests/utils/authenticateUser.ts
```diff
@@ -0,0 +1,34 @@
...
+        // setup
+        const options:OAuthUserData = {
+            email : 'x'
+        }
+        // test
+        authenticateUser({options,request:(req as any),response:(res as any)})
+        // assert
+        expect(res.status).not.toHaveBeenCalled()
+    })
+    it('should call res.status(401) if the user is not authenticated', () => {
```
SpencerKaiser, line 24:
```suggestion
    });

    it('should call res.status(401) if the user is not authenticated', () => {
```

packages/api/tests/utils/authenticateUser.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('authenticate user', () => {
+    it('should call next if the user is authenticated', () => {
+        // setup
+        const options:OAuthUserData = {
+            email : 'x'
+        }
+        // test
+        authenticateUser({options,request:(req as any),response:(res as any)})
+        // assert
+        expect(res.status).not.toHaveBeenCalled()
```
SpencerKaiser, line 22: Looks like this old; here's I'd just expect this:
``` typescript
expect(res.redirect).toHaveBeenCalledTimes(1);
expect(res.redirect).toHaveBeenCalledWith('/');
```

SpencerKaiser (changes requested):
This is awesome progress, just needs a small tweaks.

Also make sure to run `yarn lint:fix` locally to clean up the linting errors (that Action is currently failing) and patch up the tests when the rest is finished 🙂 

Great work and hope the learning journey here was helpful (sounds like it was) 🚀

packages/api/package.json
```diff
@@ -14,15 +14,17 @@
...
     "@hangar/database": "*",
     "@hangar/shared": "*",
     "@hangar/web": "*",
-    "@americanairlines/simple-env": "^1.0.4",
     "@mikro-orm/core": "^6.0.0-dev",
     "@mikro-orm/knex": "^6.0.0-dev",
     "@mikro-orm/postgresql": "^6.0.0-dev",
     "@slack/bolt": "^3.11.0",
     "@slack/web-api": "^6.6.0",
+    "@types/cookie-session": "^2.0.44",
```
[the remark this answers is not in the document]
alexanderallen-aa: yup, i was readding this morning cuz i had to rm the `yarn.lock` and i was moving too fast :)

packages/api/src/utils/authenticateUser.ts
```diff
@@ -0,0 +1,25 @@
+import { Response , Request } from 'express';
+
+type OAuthUserData = {
```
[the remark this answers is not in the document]
alexanderallen-aa: i got some ideas, but this should do for now :)

yarn.lock
```diff
@@ -3390,6 +3390,14 @@
   dependencies:
     "@types/node" "*"
 
+"@types/cookie-session@^2.0.44":
+  version "2.0.44"
+  resolved "https://packages.aa.com/artifactory/api/npm/npm-public/@types/cookie-session/-/cookie-session-2.0.44.tgz#86f3810bf19716ff716e08f877ce40cee9062267"
```
SpencerKaiser, line 3395: This is why your lint is failing; you need to remove this dependency and then re-add it after modifying your registry... if you need help with that, let me know! We might want to modify the `.npmrc` file to set the registry manually to the standard npm one

SpencerKaiser (changes requested):

SpencerKaiser (dismissed):

SpencerKaiser (approved):

PR #477, opened 2023-08-07
Branding

EatMoreOranges:
## Pre-Requisites

## Description of Changes

**Navbar:** login and sign up buttons on full size navbar are hidden once screen shrinks to smaller screens(992px or less). On smaller screens only a menu button appears for the side navbar containing the sign up and login buttons
**branding colors:** AA branding colors are mapped to color theme options. Two predefined theme options (error and errorFaint) were left alone since they did not have brand color that would be a match
**Hangar Font:** link to new [font](https://fonts.google.com/specimen/Pacifico?preview.text=Hangar&preview.text_type=custom) New font is not visible yet


packages/web/src/pages/_document.tsx
```diff
@@ -29,6 +29,12 @@ export default class Document extends NextDocument {
           <link rel="manifest" href="/manifest.json" />
           <meta name="msapplication-TileColor" content={colors.brandPrimary} />
           <meta name="msapplication-TileImage" content="/ms-icon-144x144.png" />
+          <link rel="preconnect" href="https://fonts.googleapis.com" />
```
SpencerKaiser, line 32: Can you try to follow [these docs](https://nextjs.org/docs/pages/building-your-application/optimizing/fonts) for importing fonts? I'm not sure it'll work but I don't think pulling from Google directly is the most efficient way to load fonts... if you have docs that say otherwise, please let me know!

SpencerKaiser (commented):
Small tweaks/suggestions and a question about fonts. Great start!!!

package.json
```diff
@@ -37,6 +37,7 @@
   "devDependencies": {
     "@types/cookie-session": "^2.0.44",
     "cspell": "^6.31.1",
+    "husky": "^8.0.3",
```
SpencerKaiser, line 40: We don't want husky 🙅

packages/web/package.json
```diff
@@ -14,10 +14,13 @@
     "clean": "rimraf .next dist coverage"
   },
   "dependencies": {
-    "@hangar/shared": "*",
+    "@chakra-ui/icons": "^2.1.0",
+    "@next/font": "^13.4.13",
```
SpencerKaiser, line 18: Didn't we determine this wouldn't work? if so, let's remove it

packages/web/src/components/layout/AppLayout/AppLayout.tsx
```diff
@@ -1,77 +1,42 @@
...
+        as="nav"
+        flexWrap="wrap"
+        justifyContent="space-between"
+        alignItems="center"
+        maxW={MAX_CONTENT_WIDTH_IN_PIXELS}
+        marginX="auto"
+        gap={0}
+        minH={NAV_HEIGHT}
+      >
+        {/* Add nav elements here */}
```
SpencerKaiser, line 25: This is stale, right? If so, let's remove it
```suggestion
```

SpencerKaiser (changes requested):
A few teensy things then we can merge

SpencerKaiser (approved):

PR #485, opened 2023-08-11
feat: add types for entities

alexanderallen-aa:
## Pre-Requisites


packages/shared/package.json
```diff
@@ -14,7 +14,10 @@
     "clean": "rimraf dist coverage",
     "test": "jest"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@mikro-orm/core": "^5.7.14",
```
SpencerKaiser, line 18: I don't think we need this dependency; can you remove it and rebuild everything to see if anything breaks?

packages/shared/package.json
```diff
@@ -14,7 +14,10 @@
...
+    "@hangar/database": "*"
```
SpencerKaiser, line 19: Ooooh we can get fancy with this... can you actually move this to dev dependencies? I'll make another related comment below 👀

packages/shared/src/index.ts
```diff
@@ -1,3 +1,11 @@
 export * from './config';
 export * from './types';
 export * from './utils';
+export { UserDTO,
```
SpencerKaiser, line 4: This is the related change; this should work but please let me know if there's an error on this line after making this change
```suggestion
export type { UserDTO,
```

yarn.lock
SpencerKaiser, line 1: After removing the mikro-orm dependency this should reset... _**should**_ ™️

packages/shared/src/types/entities.ts
SpencerKaiser, line 1: This all looks perfect, but I know it's going to continue growing and it's also likely we'll need to make additions to individual types that will make this file substantially larger...

Can we break this out into this kind of a setup?
```
/types
  /entities
    index.ts (individual `export * from 'ENTITY_FILE';` exports)
    User.ts (exports both User and Serialized User)
    Project
    ...
```

SpencerKaiser (changes requested):
A dependency removal, a small dependency change, and an ask to split up into multiple files. Other than that this is awesome; great job!!!

SpencerKaiser (approved):

PR #487, opened 2023-08-16
feat: user middleware

alexanderallen-aa:
## Pre-Requisites


packages/api/src/api/users/index.ts
SpencerKaiser, line 1: Can you revert your changes to this file? This middleware needs to persist as it is; creating a new user will not involve a session that has a user defined

packages/api/src/utils/addUser.ts
SpencerKaiser, line 1: can you rename as `/src/middleware/userMiddleware.ts`?

packages/api/src/utils/addUser.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const addUser = async ( req: Request, res: Response, next:Function ) => {
+  const email = req.session?.email;
+  const userQuery = { email };
+  
+  if (!email) {
+    // User does not have a valid session
+    res.sendStatus(401);
+    return;
+  }
```
SpencerKaiser, line 12: Can you utilize `id` instead? Same exact query, just leveraging the id instead of email

packages/api/src/utils/addUser.ts
```diff
@@ -0,0 +1,25 @@
...
+  
+  const user = await req.entityManager.findOne( User, userQuery );
```
SpencerKaiser, line 14: You need a `try ... catch` around this with a 500 thrown if something fails

packages/api/tests/api/users.ts/index.test.ts
SpencerKaiser, line 1: Revert the changes here; this file shouldn't change

packages/api/tests/utils/addUser.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      router.use((req, res, next) => {
+        req.session = { email: 'pancakes@waffles.bananas' } as any;
+        next();
+      }, users);
+
+      const res = await supertest(router).post('/');
+      expect(res.statusCode).toBe(200);
+    });
+  });
```
SpencerKaiser, line 50: Refactor this test to match the style of the others (no `supertest` dependency)

SpencerKaiser (changes requested):
Super close but we need to avoid changing the existing `POST /users` middleware as that needs to exist separately

granthill7 (approved):

packages/api/jest.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     '!./src/@types/**',
     '!./src/env.ts',
     '!./src/slack/*',
+    '!./src/middleware/sessionMiddleware.ts',
```
SpencerKaiser, line 17: This should not have been added here; this needs coverage

packages/api/src/middleware/sessionMiddleware.ts
```diff
@@ -3,15 +3,15 @@ import { Handler } from 'express';
...
+ * Session validity is determined by the presence of an email and id within signed cookie included with the request.
  * If the session is valid, the next function is invoked.
  *
  * @param req Express Request
  * @param res Express Response
  * @param next Express Next function
  */
 export const sessionMiddleware: Handler = (req, res, next) => {
-  if (!req.session?.email) {
+  if (!req.session?.email || !req.session?.id) {
```
SpencerKaiser, line 14: This breaks the existing functionality; now ID is required when creating a user which will always fail

SpencerKaiser (commented):
The changes to session middleware here need to be reverted

PR #489, opened 2023-08-17
feat: isolate middlewares

alexanderallen-aa:
## Pre-Requisites


packages/api/src/middleware/validateSessionMiddleware.ts
```diff
@@ -0,0 +1,14 @@
+import { Request, Response } from 'express';
+
+export const validateSession = async ( req: Request, res: Response, next:Function ) => {
+  const id = req.session?.id;
```
SpencerKaiser, line 4: This needs to be email for validateSession; userMiddleware needs to be ID

SpencerKaiser (changes requested):

PR #495, opened 2023-08-23
feat: Project Post Endpoint

alexanderallen-aa:
## Pre-Requisites


packages/api/src/api/project/post.ts
```diff
@@ -0,0 +1,36 @@
...
+  
+  const query = {id:req.user.id};
+  
+  // check for existing project
+  try {
+    const user = await em.findOne(User, query);
+    if (user?.project!==undefined) {
+      res.status(409);
+      return;
+    }
```
SpencerKaiser, line 21: Honestly we don't even need to make the query... if the `user` has a `project` at ALL, it means the entity exists and you should throw a 409...
```suggestion
    if (user?.project) {
      res.status(409);
      return;
    }
```

packages/api/src/api/project/post.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  catch {
+    res.status(500);
+    return;
+  }
+  try {
+    const project = new Project(data);
+    em.persistAndFlush(project);
```
SpencerKaiser, line 29:
```suggestion
    await em.persistAndFlush(project);
```

packages/api/src/api/project/post.ts
```diff
@@ -0,0 +1,36 @@
...
+  
+  // check for existing project
+  try {
+    const user = await em.findOne(User, query);
+    if (user?.project!==undefined) {
+      res.status(409);
+      return;
+    }
+  }
+  catch {
```
SpencerKaiser, line 23: You can get rid of the `try ... catch` after my suggestion above

packages/api/src/api/project/post.ts
```diff
@@ -0,0 +1,36 @@
...
+    return;
+  }
+  try {
+    const project = new Project(data);
+    em.persistAndFlush(project);
+    res.send(project);
+  }
+  catch(err) {
+    logger.error(err);
+    res.send(500);
```
SpencerKaiser, line 34: After creating the transaction above, you should check the error's status code and figure out if the error is an error due to the user being locked; if it is, throw a 409. [Example of how to check the status code](https://github.com/AmericanAirlines/Hangar/commit/07bfd5126b778012ed01361bbc7efa3f7e05376c#diff-c1aa0872df8ca3062b76e2375d28fa0961683b7790793fe0407ab6b113c5de8bL19-L24).

packages/api/src/api/project/post.ts
SpencerKaiser, line 29: These two lines are the ones that need to be wrapped in a transaction; you basically need to do the following:
- start a transaction
- use **_that_** entity manager to re-fetch the user but in doing so, use a `LockMode.PESSIMISTIC_WRITE_OR_FAIL` locking mode (check me on that but I'm pretty sure that's what we want)
  - If that call fails, the transaction will fail and you'll end up in your catch block; see my comment below about how to handle that
- If the user was obtainable, create the project, persist the project, then update the user with the project, persist the user, flush both

SpencerKaiser (changes requested):
You have a perfect start to it here! You're on the right path, just need to tweak how you check for a project and then add transaction support!

packages/api/src/api/project/post.ts
michael20148205: []()****@ ****

michael20148205: @

granthill7 (approved):

PR #500, opened 2023-08-29
fix: more users path cleanup

SpencerKaiser:
## Pre-Requisites


packages/api/tests/api/index.test.ts
```diff
@@ -9,7 +9,7 @@ jest.mock('../../src/api/settings', () => ({
 describe('/api route registration', () => {
   it('registers a generic catch-all handler', async () => {
     await jest.isolateModulesAsync(async () => {
-      const { api } = require('../../src/api');
+      const { api } = await import('../../src/api');
```
granthill7, line 12: Is this the async equivalent to require?

granthill7 (approved):
users bad, user good

packages/api/tests/api/index.test.ts
SpencerKaiser: They're equivalent but apparently `import` provides better context for the editor

PR #504, opened 2023-09-01
feat: Create Abstract DB entity for JudgingSession

granthill7:
## Pre-Requisites
- [ ] Yes, I updated [Authors.md](../blob/main/Authors.md) **OR** this is not my first contribution
- [ ] Yes, I wrote this code entirely myself **OR** I properly attributed these changes in [Third Party Notices](../blob/main/THIRD-PARTY-NOTICES.txt)

### DIFF
--- a/packages/database/src/entities/JudgingSession.ts
+++ b/packages/database/src/entities/JudgingSession.ts
@@ -1,0 +1,29 @@
+import { ManyToOne, Property, Entity, Ref } from '@mikro-orm/core';
+import { v4 } from 'uuid';
+import { Node } from './Node';
+import { ConstructorValues } from '../types/ConstructorValues';
+import { User } from './User';
+
+export type JudgingSessionConstructorValues = ConstructorValues<JudgingSession>;
+
+@Entity()
+export abstract class JudgingSession extends Node<JudgingSession> {
+  @Property()
+  inviteCode: string = v4();
+
+  @ManyToOne({ entity: () => User, nullable: false, ref: true })
+  createdBy: Ref<User>;
+
+  @Property()
+  createdAt: Date = new Date();
+
+  @Property({ onUpdate: () => new Date() })
+  updatedAt: Date = new Date();
+
+  constructor({ inviteCode, createdBy }: JudgingSessionConstructorValues) {
+    super();
+
+    this.inviteCode = inviteCode;
+    this.createdBy = createdBy;
+  }
+}
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -30,12 +30,14 @@
     "@mikro-orm/sql-highlighter": "^1.0.1",
     "dotenv-flow": "^3.2.0",
     "mikro-orm": "^6.0.0-dev",
-    "readline-sync": "^1.4.10"
+    "readline-sync": "^1.4.10",
+    "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@types/jest": "^29.2.1",
     "@types/node": "^16.9.1",
     "@types/readline-sync": "^1.4.4",
+    "@types/uuid": "^9.0.3",
     "@typescript-eslint/eslint-plugin": "^5.48.1",
     "@typescript-eslint/parser": "^5.48.1",
     "dotenv-flow": "^3.2.0",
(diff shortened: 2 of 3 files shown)


packages/database/src/entities/JudgingSession.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ConstructorValues } from '../types/ConstructorValues';
+import { User } from './User';
+
+export type JudgingSessionDTO = EntityDTO<JudgingSession>;
+
+export type JudgingSessionConstructorValues = ConstructorValues<JudgingSession>;
+
+@Entity()
+export abstract class JudgingSession extends Node<JudgingSession> {
+  @Property()
```
SpencerKaiser, line 13:
```suggestion
  @Property({ unique: true })
```

packages/database/src/entities/ExpoJudgingSession.ts
```diff
@@ -0,0 +1,5 @@
+import { Entity } from '@mikro-orm/core';
+import { JudgingSession } from './JudgingSession';
+
+@Entity()
+export class ExpoJudgingSession extends JudgingSession {}
```
SpencerKaiser, line 5: We need the DTO for this too

packages/database/src/entities/JudgingSession.ts
```diff
@@ -0,0 +1,31 @@
...
+  inviteCode: string = v4();
+
+  @ManyToOne({ entity: () => User, nullable: false, ref: true })
+  createdBy: Ref<User>;
+
+  @Property()
+  createdAt: Date = new Date();
+
+  @Property({ onUpdate: () => new Date() })
+  updatedAt: Date = new Date();
```
SpencerKaiser, line 23: You'll get these properties by using `Node`
```suggestion
```

packages/database/src/entities/JudgingSession.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor({ inviteCode, createdBy }: JudgingSessionConstructorValues) {
+    super();
+
+    this.inviteCode = inviteCode;
```
SpencerKaiser, line 28: Invite code is set automatically so it needs to be excluded from the constructor
```suggestion
  constructor({ createdBy }: JudgingSessionConstructorValues) {
    super();

```

packages/database/src/entities/JudgingSession.ts
```diff
@@ -0,0 +1,31 @@
+import { ManyToOne, Property, Entity, Ref, EntityDTO } from '@mikro-orm/core';
+import { v4 } from 'uuid';
+import { Node } from './Node';
+import { ConstructorValues } from '../types/ConstructorValues';
+import { User } from './User';
+
+export type JudgingSessionDTO = EntityDTO<JudgingSession>;
+
+export type JudgingSessionConstructorValues = ConstructorValues<JudgingSession>;
```
SpencerKaiser, line 9: See my comments on the constructor
```suggestion
export type JudgingSessionConstructorValues = ConstructorValues<JudgingSession, 'inviteCode'>;
```

SpencerKaiser (changes requested):

packages/database/src/migrations/0000-add-initial-entities.ts
SpencerKaiser, line 1: Both this file and the file below need to be reverted 😬 when you asked if modifying migrations was ok I thought you were referring to net-new migrations... migrations that have already been committed should never be modified because it'll brick environments (changes to old migrations will never be run unless the DB is torn down entirely and spun back up again)

After renaming this entity the ORM should pick up these changes and include them as part of your migration; if that's _**not**_ the case, please let me know and I'll work with you to write any needed SQL statements manually

SpencerKaiser (changes requested):
The migration changes I mention in the review are gonna be a little rough... here's what I suggest: 

- Run the `down` command locally 
- Force pull the entire `migrations` directory into your local and then re-create your migrations from scratch. If the changes to `JudgingVote` aren't made automatically, ping me and we can add those manually

Also make sure to address the comments I made in my last review

SpencerKaiser (approved):

PR #505, opened 2023-09-05
feat: adding better Slack docs, cleaning up some schema, adding Node utility type to Shared

SpencerKaiser:
## Pre-Requisites

## Description of Changes

👆


packages/shared/src/schema/user/put.ts
```diff
@@ -1,6 +1,15 @@
 import { z } from 'zod';
 
-const nameSchema = z.string().min(2).max(50);
+export const PutValidation = {
```
granthill7, line 3: Is this a better way to do this or just preference?

packages/shared/src/types/utilities/Pretty.ts
```diff
@@ -0,0 +1,3 @@
+export type Pretty<K> = {
```
granthill7, line 1: Can you explain to me what the benefit of this type is?

granthill7 (approved):

packages/shared/src/types/utilities/Pretty.ts
SpencerKaiser: It flattens the object so the type that you see in the editor is human-friendly instead of a nasty union of several types 😬

packages/shared/src/schema/user/put.ts
SpencerKaiser: The UI will need the values to help guide the user _**before**_ a validation failure occurs... do you have a better idea for naming? I struggled with this a bit and just settled on this pattern but happy to change it if you have a better recc.

PR #510, opened 2023-09-07
feat: Events list UI

alexanderallen-aa:
## Pre-Requisites


packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Event, eventSchema, EventsList } from '../components/EventsList';
+import { PageContainer } from '../components/layout/PageContainer';
+
+const fetchEvents = async () => {
+  const res = await fetch('/api/event');
```
SpencerKaiser, line 6:
```suggestion
  const res = await axios.get<SerializedEvent>('/api/event');
```

packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Event, eventSchema, EventsList } from '../components/EventsList';
+import { PageContainer } from '../components/layout/PageContainer';
+
+const fetchEvents = async () => {
```
SpencerKaiser, line 5:
```suggestion
const fetchEvents: () => Promise<Event[]> = async () => {
```

packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Event, eventSchema, EventsList } from '../components/EventsList';
```
SpencerKaiser, line 2:
```suggestion
import { Event, SerializedEvent } from '@hangar/shared';
```

packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <div>
+      <PageContainer pageTitle={'Schedule'} heading={'Events'}>
+        {/* cspell:disable-next */}
+      </PageContainer>
+      <div style={{width:'350px'}}>
+        <EventsList {...{events}} />
+      </div>
+    </div>
+  );
```
SpencerKaiser, line 34:
```suggestion
  return (
      <PageContainer pageTitle={'Schedule'} heading={'Events'}>
        <EventsList events={events} />
      </PageContainer>
  );
```

packages/web/src/components/EventsList/utils.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+  
+export const eventSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  description: z.string(),
+  start: z.string(),
+  end: z.string(),
+});
+
```
SpencerKaiser, line 17:
```suggestion
```

packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const fetchEvents = async () => {
+  const res = await fetch('/api/event');
+
+  if (res.ok) {
+    const data: Event[] = await res.json();
+    const events: Event[] = (data).map((event) => eventSchema.parse(event));
+    
+    return events;
+  }
```
SpencerKaiser, line 13: With axios the result should be something like this:
```suggestion
  const { start, end, ...rest } = res;
  return data.map((serializedEvent) => {...rest, start: dayjs(start), end: dayjs(end) });
```

End result is an array of events that have dayjs objects so we won't need to use `Date` functionality anywhere

packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const Schedule: React.FC = () => {
```
SpencerKaiser, line 16:
```suggestion
const Schedule: NextPage = () => {
```

packages/web/src/pages/schedule.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [events, setEvents] = React.useState<Event[]>([]);
+  
+  React.useEffect(() => {
+    void fetchEvents().then( (events) => {
+      setEvents(events as Event[]);
+    });
```
SpencerKaiser, line 22:
```suggestion
    const fetchAndSetEvents = async () => {
      try {
        setEvents(await fetchEvents());
      } catch (error) {
        // eslint-disable-next-line no-console
        console.error('Failed to fetch events');
        // TODO: Show error toast on failure        
      }
    };
    
    void fetchAndSetEvents();
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+        ? <RegistrationModal />
+        : ''
+      } */}
+    
+    </li>
+  );
+}
+  
+export const EventsList: React.FC<{ events: Event[] }> = ({ events }) => (
+  <ul>
```
SpencerKaiser, line 52: Instead of `ul` here, try `<Flex direction="column">`

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  
+  const now = new Date();
+  const _start = new Date(start);
+  const _end = new Date(end);
+  const inProgress = _start < now && now < _end;
+  const past = now > _end;
+  const badge = inProgress ? 'IN PROGRESS' : past ? 'PAST' : '';
+  
+  return (
+    <li style={eventStyle(badge)}>
```
SpencerKaiser, line 20: Instead of `li` here, try `<Flex direction="column">`

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <h2>
```
SpencerKaiser, line 21: `<Heading>`

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+        {name}&nbsp;&nbsp;
+        
+        { (badge=='IN PROGRESS') &&
+          <Badge colorScheme="green" variant="solid" style={{ float:'right' }}>
+            {badge} &nbsp;&nbsp;
+            {/* clock icon */}
+            &#x23F0;
```
SpencerKaiser, line 28: This is perfect for `<Flex direction="row" gap={2}>` and just use the straight emoji instead of the unicode; it renders the same and is more readable in code

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+          </Badge>
+        }
+      
+      </h2>
+      
+      <p>
```
SpencerKaiser, line 34: `<Text>`

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+        {description}
+      </p>
+      
+      <p>
```
SpencerKaiser, line 38: Same ^

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { Event, eventStyle } from './utils';
+import { Badge } from '@chakra-ui/react';
+
+const Event = ({ event }: { event: Event }) => {
+  
+  const { name , description, start, end } = event;
+  const stringFormat: Intl.DateTimeFormatOptions = { hour: 'numeric', minute: 'numeric' };
```
SpencerKaiser, line 8:
```suggestion
  const stringFormat = 'HH:mm a'; // e.g., 12:45 pm
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const formattedStart = new Date(start).toLocaleTimeString([], stringFormat);
+  const formattedEnd = new Date(end).toLocaleTimeString([], stringFormat);
```
SpencerKaiser, line 10:
```suggestion
  const formattedStart = start.format(stringFormat);
  const formattedEnd = end.format(stringFormat);
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  
+  const now = new Date();
+  const _start = new Date(start);
+  const _end = new Date(end);
```
SpencerKaiser, line 14:
```suggestion
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const inProgress = _start < now && now < _end;
```
SpencerKaiser, line 15:
```suggestion
  const inProgress = dayjs().isBetween(start, end);
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const past = now > _end;
```
SpencerKaiser, line 16:
```suggestion
  const past = dayjs().isAfter(end);
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { Event, eventStyle } from './utils';
+import { Badge } from '@chakra-ui/react';
+
```
SpencerKaiser, line 4:
```suggestion

type EventProps = {
  event: Event
};

```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { Event, eventStyle } from './utils';
```
SpencerKaiser, line 2:
```suggestion
import { eventStyle } from './utils';
import { Event } from '@hangar/shared';
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Badge } from '@chakra-ui/react';
+
+const Event = ({ event }: { event: Event }) => {
```
SpencerKaiser, line 5:
```suggestion
const Event: React.FC<EventProps> = ({ event }) => {
```

packages/web/src/components/EventsList/utils.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const eventStyle =  (badge:string) => ({
+  color: badge!=='PAST'?'':'#666',
+  padding: '8px' ,
+  margin: '8px' ,
+  listStyleType: 'none',
+  border: '1px solid #444',
+  borderRadius:'8px',
+  background:badge!=='PAST'?'#222':'#111',
```
SpencerKaiser, line 25: For colors, make sure to pull from `colors.ts`

SpencerKaiser (changes requested):
Excellent structure! Just a couple of big things to consider as you continue to work:
- In the UI please use `Axios` for http requests; the TL;DR on why is better types and better SSR support should we decide to swap to SSR for any of these pages
- We shouldn't need raw html anywhere; please use Chakra-UI for your components (suggestions on which components are relevant can be found in inline comments)

packages/web/src/components/EventsList/index.tsx
```diff
@@ -0,0 +1,2 @@
+export * from './EventsList';
+export * from './utils';
```
SpencerKaiser, line 2: This shouldn't be necessary outside this component; if it is, let's move the relevant pieces to a separate file that's more globally accessible.

```suggestion
```

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <Heading as="h2" size="md">
+        {name}
+      </Heading>
+      
+      <Text>
+        {description}
+      </Text>
+      
+      <Text>
+        {start.format('HH:mm a')} - {end.format('HH:mm a')}
```
SpencerKaiser, line 63: Can you split the format string out into a var that's used in both spots?

packages/web/src/components/EventsList/utils.tsx
```diff
@@ -0,0 +1,20 @@
+import { colors } from '../../theme/colors'
+
```
SpencerKaiser, line 2: Let's do something like this; I didn't realize it was being used across files too... this file and the `EventList` should use the same values so it doesn't break if one changes.
```suggestion

export type EventStatus = 'IN PROGRESS' | 'PAST' | 'FUTURE';

```

Feel free to use an enum if you feel that would be more appropriate.

SpencerKaiser (changes requested):
Tiny tweaks and then we're good to go

SpencerKaiser (dismissed):

SpencerKaiser (approved):

PR #513, opened 2023-09-08
feat: Add repoLink to POST Schema

Ansaar-Syed:
## Pre-Requisites


packages/api/src/api/project/post.ts
```diff
@@ -25,6 +26,10 @@ export const post = async (req: Request, res: Response) => {
 
       project = new Project(data);
       project.contributors.add(lockedUser);
+      const hostRes = await axios.get(data.repoLink);
```
SpencerKaiser, line 29: This is unrelated to the line above so let's break it up a bit
```suggestion

      const hostRes = await axios.get(data.repoLink);
```

Also what is `hostRes`? Are you fetching a host? – please rename to something that's a bit more descriptive like `repoFetchRes`

Finally you need a `try ... catch` around this fetch

packages/api/src/api/project/post.ts
```diff
@@ -25,6 +26,10 @@ export const post = async (req: Request, res: Response) => {
...
+      if (hostRes.status !== 200) {
+        throw new Error('oh no repoLink cannot be found');
```
SpencerKaiser, line 31: This will simply break the request and cause a timeout on behalf of the client; this needs to be a structured response to the front end with a meaningful error

```suggestion
        res.status(400).send('Repo URL could not be validated');
        return;
```

packages/database/src/entities/Project.ts
```diff
@@ -23,6 +23,9 @@ export class Project extends Node<Project> {
   @Property({ columnType: 'text' })
   name: string;
 
+  @Property({ columnType: 'text', nullable: true })
```
SpencerKaiser, line 26: AC didn't say this was supposed to be optional 😬 Your schema also has it as required

```suggestion
  @Property({ columnType: 'text' })
```

packages/database/src/migrations/0005-add-repolink-to-project.ts
SpencerKaiser, line 1: You'll have to recreate the migration after the changes above

packages/database/src/entities/Project.ts
```diff
@@ -23,6 +23,9 @@ export class Project extends Node<Project> {
...
+  repoLink: string;
```
SpencerKaiser, line 27: > _In everyday speech, “URL” and “link” are often used interchangeably. Strictly speaking, however, they are different – a link describes the function, which is to take you from one place on the internet to another, while the URL describes the location itself, the place you want to go_

AC mentioned URL

```suggestion
  repoUrl: string;
```

packages/database/src/migrations/.snapshot-hangar.json
SpencerKaiser, line 1: You'll have to recreate the migration after the changes above

packages/shared/src/schema/project/post.ts
```diff
@@ -16,4 +16,8 @@ export const post = z.object({
     .min(PostValidation.MIN_DESCRIPTION_LENGTH)
     .max(PostValidation.MAX_DESCRIPTION_LENGTH),
   location: z.string().trim().max(PostValidation.MAX_LOCATION_LENGTH).optional(),
+  repoLink: z
```
SpencerKaiser, line 19: Please include test changes to address this addition

packages/api/src/api/project/post.ts
SpencerKaiser, line 1: Tests need to be added to address these changes

SpencerKaiser (changes requested):

packages/database/src/entities/Project.ts
granthill7: Do we not need to leave the option open for a project to not have a repoLink right off of the bat? This was my call

packages/api/src/api/project/post.ts
granthill7: So even inside of the try block we should just immediately send the response instead of throwing to the catch block?

packages/shared/src/schema/project/post.ts
```diff
@@ -16,4 +16,8 @@
...
+    .string()
+    .includes('github.com' || 'bitbucket.com' || 'gitlabs.com')
```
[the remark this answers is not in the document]
SpencerKaiser: @Ansaar-Syed @granthill7 please address the above ^ there might be a way to check that the URL starts with `https://URL`

packages/api/src/api/project/post.ts
SpencerKaiser: Oh I totally missed that... ok the way yall are doing it is fine but the error message is way too playful for a server; can you sub mine for what you have?

packages/api/src/api/project/post.ts
granthill7: A try inside of a try?

packages/database/src/entities/Project.ts
SpencerKaiser: Right now we have no `PUT` endpoint so the project creation needs to be final

packages/api/src/api/project/post.ts
SpencerKaiser: Yep good call on the try; the name of the variable still needs to change though

packages/shared/src/schema/project/post.ts
```diff
@@ -16,4 +16,8 @@
     .min(PostValidation.MIN_DESCRIPTION_LENGTH)
     .max(PostValidation.MAX_DESCRIPTION_LENGTH),
   location: z.string().trim().max(PostValidation.MAX_LOCATION_LENGTH).optional(),
+  repoUrl: z
+    .string()
+    .includes('github.com' || 'bitbucket.com' || 'gitlabs.com')
```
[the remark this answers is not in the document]
SpencerKaiser: This is still valid; please review and ask questions if it's unclear what the expectation is

SpencerKaiser (changes requested):
Several comments haven't been addressed yet

packages/shared/tests/schema/project/post.test.ts
```diff
@@ -76,4 +77,8 @@ describe('project post schema', () => {
       }).success,
     ).toBe(false);
   });
+
+  it('validates a matching RepoUrl', () => {
+    expect(Schema.project.post.safeParse(validProject.repoURl).success).toBe(true);
```
SpencerKaiser, line 82: You need to pass a full object to this; look at the tests above for examples. e.g., if you want to override the `description` from the `validProject`, this is what you'd do:
``` typescript
{
    ...validProject,
    description: Array(PostValidation.MIN_DESCRIPTION_LENGTH - 1).fill('A'),
  }
```
you'll need to do something similar and then pass in a good repo url and a bad one and make sure both work as expected

packages/shared/src/schema/project/post.ts
```diff
@@ -16,4 +16,15 @@ export const post = z.object({
...
     .max(PostValidation.MAX_DESCRIPTION_LENGTH),
   location: z.string().trim().max(PostValidation.MAX_LOCATION_LENGTH).optional(),
+  repoUrl: z
+    .string()
+    .trim()
+    .url()
+    .refine(
+      (url) =>
+        url.startsWith('https://github.com/') ||
+        url.startsWith('https://bitbucket.com/') ||
```
SpencerKaiser, line 26: After some research I think it's best we exclude this as an option
```suggestion
```

SpencerKaiser (changes requested):

SpencerKaiser (dismissed):

packages/shared/src/schema/project/post.ts
```diff
@@ -16,4 +16,14 @@ export const post = z.object({
...
     .max(PostValidation.MAX_DESCRIPTION_LENGTH),
   location: z.string().trim().max(PostValidation.MAX_LOCATION_LENGTH).optional(),
+  repoUrl: z
+    .string()
+    .trim()
+    .url()
+    .refine(
+      (url) =>
+        url.startsWith('https://github.com/') ||
+        url.startsWith('https://gitlabs.com/'),
```
SpencerKaiser, line 26:
```suggestion
      (url) => url.startsWith('https://github.com/') || url.startsWith('https://gitlabs.com/'),
```

SpencerKaiser (approved):

PR #514, opened 2023-09-08
feat: ExpoJudingSession frontend store

granthill7:
## Pre-Requisites


packages/web/src/stores/expoJudgingSession.ts
```diff
@@ -0,0 +1,53 @@
...
+        console.error(error);
+        // TODO: Show error toast
+      }
+    }
+    set({ expoJudgingSessions, doneLoading: true });
+  },
+  addExpoJudgingSessions: async (esj: ExpoJudgingSession) => {
+    try {
+      set((state) => ({
+        expoJudgingSessions: state.expoJudgingSessions?.concat([esj]),
```
SpencerKaiser, line 42:
```suggestion
        expoJudgingSessions: [...state.expoJudgingSessions, esj],
```

packages/web/src/stores/expoJudgingSession.ts
```diff
@@ -0,0 +1,53 @@
...
+  fetchExpoJudgingSessions: async () => {
+    let expoJudgingSessions: ExpoJudgingSession[] | undefined;
+    try {
+      const res = await axios.get<SerializedExpoJudgingSession[]>(`/api/expoJudgingSession`);
+      expoJudgingSessions = res.data.map(
+        (serializedEJS) =>
+          ({
+            id: serializedEJS.id,
+            inviteCode: serializedEJS.inviteCode,
+            createdBy: serializedEJS.createdBy,
```
SpencerKaiser, line 25:
```suggestion
        ({ createdAt, UpdatedAt, ...rest) =>
          ({
            ...rest,
```

packages/web/src/stores/expoJudgingSession.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (!axios.isAxiosError(error) || error.status !== 401) {
+        // eslint-disable-next-line no-console
+        console.error(error);
+        // TODO: Show error toast
+      }
+    }
+    set({ expoJudgingSessions, doneLoading: true });
+  },
+  addExpoJudgingSessions: async (esj: ExpoJudgingSession) => {
+    try {
```
SpencerKaiser, line 40: You don't need a `try ... catch` here; if the thing calling this has bad data it should be a compile time error

SpencerKaiser (changes requested):

packages/web/src/stores/expoJudgingSession.ts
SpencerKaiser: It's also worth calling out that your original code would have just set it to undefined if the original object hadn't been initialized; always be mindful of what happens if the optional thing there is undefined

PR #516, opened 2023-09-12
feat: Prizes list UI

alexanderallen-aa:
## Pre-Requisites

### DIFF
--- a/packages/web/src/components/EventsList/EventsList.tsx
+++ b/packages/web/src/components/EventsList/EventsList.tsx
@@ -78,7 +78,7 @@ const EventCard: React.FC<EventCardProps> = ({ event }) => {
 export const EventsList: React.FC<{ events: Event[] }> = ({ events }) => (
   <UnorderedList variant="card">
     {events.map((event) => (
-      <EventCard {...{ event, key: event.id }} key={`event-${event.id}`} />
+      <EventCard {...{ event }} key={`event-${event.id}`} />
     ))}
   </UnorderedList>
 );
--- a/packages/web/src/components/Prizes/Prizes.tsx
+++ b/packages/web/src/components/Prizes/Prizes.tsx
@@ -14,15 +14,14 @@ const FAILED_FETCH_TOAST: UseToastOptions = {
 };
 
 const fetchPrizes: () => Promise<Prize[]> = async () =>
-  (await axios.get<SerializedPrize[]>('/api/prize'))
-    .data.map((serializedPrize) => {
-      const { createdAt, updatedAt, ...rest } = serializedPrize;
-      return {
-        ...rest,
-        createdAt: dayjs(createdAt),
-        updatedAt: dayjs(updatedAt),
-      };
-    } );
+  (await axios.get<SerializedPrize[]>('/api/prize')).data.map((serializedPrize) => {
+    const { createdAt, updatedAt, ...rest } = serializedPrize;
+    return {
+      ...rest,
+      createdAt: dayjs(createdAt),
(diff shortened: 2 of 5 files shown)


packages/web/src/components/Prizes/PrizesList.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type PrizesListProps = {
+  prizes: Prize[];
+};
+
+const RANKING= [ '🥇', '🥈', '🥉' ];
+const BONUS = '✨';
+
+const PrizeCard: React.FC<PrizeCardProps> = ({ prize }) => {
+  const { name, description, position, isBonus } = prize;
```
granthill7, line 18: position will be in [0, 1, 2] correct?

packages/web/src/components/Prizes/Prizes.tsx
```diff
@@ -0,0 +1,46 @@
...
+    } );
+
+export const Prizes: React.FC = () => {
+  const [prizes, setPrizes] = React.useState<Prize[]>([]);
+  const toast = useToast();
+
+  React.useEffect(() => {
+    const fetchAndSetPrizes = async () => {
+      try {
+        setPrizes(await fetchPrizes());
```
granthill7, line 34: Do we need a success toast or is that out of scope?

alexanderallen-aa: not sure, the failure was a recommendation from spencer, but there was no mention of success...

packages/web/src/components/Prizes/PrizesList.tsx
alexanderallen-aa: it could be any positive int as far I can tell, but it is only used here when -1>position<3 per line 27

packages/web/src/components/EventsList/EventsList.tsx
```diff
@@ -78,7 +78,7 @@ const EventCard: React.FC<EventCardProps> = ({ event }) => {
 export const EventsList: React.FC<{ events: Event[] }> = ({ events }) => (
   <UnorderedList variant="card">
     {events.map((event) => (
-      <EventCard {...{ event, key: event.id }} key={`event-${event.id}`} />
+      <EventCard {...{ event }} key={`event-${event.id}`} />
```
SpencerKaiser, line 81: Good catch; the linter was unhappy with the other approach as it didn't find it for some reason

packages/web/src/components/Prizes/Prizes.tsx
SpencerKaiser: Yea no success toast; success is displaying the results

packages/web/src/components/Prizes/PrizesList.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <ListItem background={colors.brandPrimaryDark}>
+      <Heading as="h2" size="md">
+        <Flex gap={2}>
+          <Text>
+            { isBonus // eslint-disable-line 
+              ? BONUS
+              : position < RANKING.length
+                ? RANKING[position]
+                : null
+            }
```
SpencerKaiser, line 30: This is a little cleaner, but also please don't nest ternary expressions; this eslint rule is a good one
```suggestion
            { isBonus ? BONUS : RANKING[position + 1] ?? BONUS }
```

SpencerKaiser (changes requested):
One teensy change and we can ship

EatMoreOranges (approved):

PR #520, opened 2023-09-14
feat: project registration ui

alexanderallen-aa:
## Pre-Requisites


SpencerKaiser (changes requested):
Fantastic start, we just need to reduce/simply a bit...

The shared modal to do both create/edit flows was a bad call when we first built that on v2 and I definitely don't want us to follow in those footsteps. In addition, the `PUT` vs `POST` stuff was out of scope for this US (and may be out of scope for MVP). On your next pass, focus on reducing complexity and removing stuff that isn't MVP; simple/small/readable is the goal.


Re: Validation - the below clip is very weird behavior; I expect validation errors to only pop up _once a save attempt as been made_. I'd be ok if the error showed for a field once the field had been touched and then loses focus but it's a bug if another, unrelated field showing is showing a validation error if it hasn't been touched.

packages/web/src/components/PopupModal/PopupModal.tsx
SpencerKaiser, line 1: Can you move this into the `RegistrationModal` directory? I don't see us using this for other purposes as we'll almost certainly create individual modals in the future instead of creating a wrapper around them.

packages/web/src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { popupProps } from './utils';
+
+interface PopupModalProps {
+  openModalText: String;
+  header: String;
+  hideCloseButton?: boolean;
+  openButtonProps?: ButtonProps;
+  onConfirm?: () => Promise<void>;
+  succussMessage?: string;
+  errorMessage?: string;
```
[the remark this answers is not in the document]
SpencerKaiser: This wasn't resolved

packages/web/src/components/PopupModal/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  setIsLoading,
+  onConfirm,
+  setModalHeader,
+  setBodyText,
+  onClose,
+  setConfirmButtonVisible,
+  onOpen,
+  isLoading,
+  succussMessage,
+  errorMessage,
```
SpencerKaiser, line 28: See my unresolved comment on these; neither is being used so there's a bunch of code below we can rip out

packages/web/src/components/PopupModal/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+      setModalHeader(header);
+      setBodyText(undefined);
+      setConfirmButtonVisible(true);
+      onOpen();
+    },
+  };
+
+  const confirmProps = {
+    isLoading,
+    colorScheme: 'blue',
```
SpencerKaiser, line 64: This blue isn't being used anywhere else so let's leave this to default colors/theme
```suggestion
```

packages/web/src/components/PopupModal/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  onConfirm?: () => Promise<void>;
+  setModalHeader: React.Dispatch<React.SetStateAction<string>>;
+  setBodyText: React.Dispatch<React.SetStateAction<string | undefined>>;
+  onClose: () => void;
+  setConfirmButtonVisible: React.Dispatch<React.SetStateAction<boolean>>;
+  onOpen: () => void;
+  isLoading: boolean;
+  succussMessage?: string;
+  errorMessage?: string;
+  openButtonPropsLocal: any;
```
SpencerKaiser, line 14: Delete this per my comment above but also watch out for the `any` next time!

packages/web/src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ModalHeader,
+  ModalOverlay,
+  useDisclosure,
+} from '@chakra-ui/react';
+import { popupProps } from './utils';
+
+interface PopupModalProps extends PropsWithChildren {
+  openModalText: string;
+  header: string;
+  hideCloseButton?: boolean;
```
SpencerKaiser, line 20: Remove this; it isn't being used

packages/web/src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  openButtonProps?: ButtonProps;
```
SpencerKaiser, line 21: This value is being passed around all over the place and I don't think we should even be supporting this; let's scrap both the original and the alias's usage. If you want the button in the modal to pop, use `variant="cta"`.

packages/web/src/components/PopupModal/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  isLoading,
+  succussMessage,
+  errorMessage,
+  openButtonPropsLocal,
+  header,
+}: PopupProps) => {
+  const onConfirmClick = async () => {
+    try {
+      setIsLoading(true);
+      await onConfirm!();
```
SpencerKaiser, line 35: We can't assert this exists; if it needs to exist, change the types to make it required instead of optional
```suggestion
      await onConfirm();
```

packages/web/src/components/RegistrationModal/useRegistrationConfig.tsx
```diff
@@ -0,0 +1,73 @@
...
+        }
+      }
+      return errors;
+    },
+    async onSubmit(values) {
+      const res = await fetch(`/api/project/${initialValues?.id ?? ''}`, {
+        method: initialValues ? 'PUT' : 'POST',
+        body: JSON.stringify({
+          ...values,
+          location: values.location!.trim(),
```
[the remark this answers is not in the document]
SpencerKaiser: Not resolved

packages/web/src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  onConfirm?: () => Promise<void>;
```
SpencerKaiser, line 22: Am I crazy or is this unused as well? I don't see a single place where an `onConfirm` is passed as a prop... if it's unused, please delete that as well.

packages/web/src/components/RegistrationModal/RegistrationForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+    ...Object.entries(formik.errors)
+      .map(([key]) => {
+        const value = formik.values[key as keyof RegistrationSchema];
+        return { [key]: value };
+      })
+      .reduce((acc, curr) => ({ ...acc, ...curr }), {}),
+  };
+  await formik.validateForm(values);
+};
+const formProps: FormProps = (key, formik) => ({
```
SpencerKaiser, line 33:
```suggestion
};

const formProps: FormProps = (key, formik) => ({
```

packages/web/src/components/RegistrationModal/RegistrationForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+  return (
+    <Box>
+      <form onSubmit={formik.handleSubmit}>
+        <VStack alignItems="stretch">
+          <FormControl id="name">
+            <FormLabel>
+              Name
+              <Hint> A descriptive title for your app </Hint>
+            </FormLabel>
+            <Input {...(formProps('name', formik) as InputProps)} />
```
SpencerKaiser, line 56: Was the assertion here needed? If so let's chat because a better way to do this would be to add a generic to the type for `formProps`

packages/web/src/components/RegistrationModal/HintTooltip.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Box, Tooltip } from '@chakra-ui/react';
+import { colors } from '../../theme/colors';
+
+export const Hint: React.FC<{ children: string }> = ({ children }) => (
```
SpencerKaiser, line 5: File name is inconsistent with the export; please make sure the two are identical but I don't have a preference as to which one

Also the type on this `children` prop is incorrect; it CAN be a string but is not guaranteed to be one. It also has a separate type than the type of `children` everywhere else.
```suggestion
type HintProps = PropsWithChildren;

export const Hint: React.FC<HintProps> = ({ children }) => (
```

packages/web/src/components/RegistrationModal/RegistrationForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onBlur: (e: React.FocusEvent<HTMLInputElement>) => handleBlur(e, formik),
+});
+
+export const RegistrationForm: React.FC<RegistrationFormProps> = ({ initialValues, onSubmit }) => {
+  const { formik, alertDescription, alertProps } = useRegistrationConfig({
+    initialValues,
+    onSubmit,
+  });
+  return (
+    <Box>
```
SpencerKaiser, line 48: Is this `Box` necessary? 🤔

packages/web/src/components/RegistrationModal/useRegistrationConfig.tsx
```diff
@@ -0,0 +1,67 @@
...
+          if (issue.path && (issue.path[0] as string) in values) {
+            errors[issue.path[0] as keyof RegistrationSchema] = issue.message;
+          }
+        });
+      }
+
+      return errors;
+    },
+    async onSubmit(values) {
+      const res = await fetch(`/api/project/`, {
```
SpencerKaiser, line 35: Please use axios
```suggestion
      const res = await fetch(`/api/project`, {
```

SpencerKaiser (changes requested):

packages/web/src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ButtonProps,
+  HStack,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  useDisclosure,
```
SpencerKaiser, line 13: Test?

packages/web/src/components/PopupModal/PopupModal.tsx
```diff
@@ -0,0 +1,79 @@
...
+  Button,
+  ButtonProps,
+  HStack,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
```
SpencerKaiser, line 12: TEST?
```suggestion
  ModalContent,
  ModalFooter,
  ModalHeader,
  ModalOverlay,
```

packages/web/src/components/RegistrationModal/HintTooltip.tsx
SpencerKaiser, line 1: Please make sure the file and the export have the same name

packages/web/src/components/RegistrationModal/RegistrationForm.tsx
SpencerKaiser: In the future we can look at generics

packages/web/src/components/RegistrationModal/RegistrationForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+  await formik.validateForm(values);
+};
+const formProps: FormProps = (key, formik) => ({
+  variant: 'filled',
+  type: 'text',
+  value: formik.values[key],
+  isInvalid: !!formik.errors[key],
+  onChange: formik.handleChange,
+  onBlur: (e: React.FocusEvent<HTMLInputElement>) => handleBlur(e, formik),
+});
```
SpencerKaiser, line 40: Try this for your types issues
```suggestion
} as const);
```

SpencerKaiser (approved):

PR #523, opened 2023-09-16
feat: adding logout and fixing admin store fetch

SpencerKaiser:
## Pre-Requisites

### DIFF
--- a/packages/api/src/utils/authenticateUser.ts
+++ b/packages/api/src/utils/authenticateUser.ts
@@ -24,11 +24,11 @@ export const authenticateUser = async ({ data, req, res }: AuthenticateArgs) => 
   try {
     const existingUser = await em.findOne(User, { email: data.email });
     if (existingUser) {
-      req.session.id = existingUser.id;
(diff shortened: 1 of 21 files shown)


packages/api/src/utils/authenticateUser.ts
```diff
@@ -24,11 +24,11 @@ export const authenticateUser = async ({ data, req, res }: AuthenticateArgs) =>
   try {
     const existingUser = await em.findOne(User, { email: data.email });
     if (existingUser) {
-      req.session.id = existingUser.id;
+      req.session = { id: existingUser.id };
```
granthill7, line 27: Is there a reason this is better? Or just best practice?

SpencerKaiser: The change in types I made makes session either a `Session` or `null`; as a result, I would have had to do `req.session?.id =` and you can't assign that way: 

image

The change in type was necessary because our cookie library doesn't like the removal of that id field and instead prefers that the entire session be set to `null` in order to clear it

PR #540, opened 2023-10-02
fix: fixing erroneous toasts, fixing deprecations

SpencerKaiser:
## Pre-Requisites


packages/web/src/stores/admin.ts
```diff
@@ -20,7 +20,7 @@ export const useAdminStore = create<AdminStore>((set) => ({
       const { createdAt, updatedAt, ...rest } = res.data;
       admin = { ...rest, createdAt: dayjs(createdAt), updatedAt: dayjs(updatedAt) };
     } catch (error) {
-      if (!axios.isAxiosError(error) || error.status !== 401) {
+      if (!axios.isAxiosError(error) || ![401, 403].includes(error.response?.status as number)) {
```
Ansaar-Syed, line 23: can you explain me what you did here ? @SpencerKaiser

SpencerKaiser: Changed the check to see if the error is a 403 or a 401 instead of just a 401... we really only want this to pop up when it's a 404 or a 500

Ansaar-Syed: Got it thanks

Ansaar-Syed (approved):

PR #542, opened 2023-10-02
feat: Adding Expo Judging page with invite code support

SpencerKaiser:
## Pre-Requisites


packages/web/src/stores/user.ts
```diff
@@ -31,5 +32,8 @@ export const useUserStore = create<UserStore>((set) => ({
       }
     }
     set({ user, doneLoading: true });
+
+    // Load user-dependent stores
+    void useAdminStore.getState().fetchAdmin();
```
alexanderallen-aa, line 37: what does this void do in this context?

SpencerKaiser: Basically says "don't wait on this admin fetch to complete; just trigger it and move on"

SpencerKaiser: It's the same as leaving nothing but `eslint` wants us to pick a lane so we don't forget to await a promise

alexanderallen-aa (approved):

PR #544, opened 2023-10-02
fix: fixing issues related to the new ExpoJudgingSessionContext entity

SpencerKaiser:
## Pre-Requisites


packages/database/src/entities/ExpoJudgingSessionContext.ts
```diff
@@ -10,7 +10,7 @@ type ConstructorArgs = Pick<ExpoJudgingSessionContext, 'judge' | 'expoJudgingSes
 @Entity()
 @Unique({ properties: ['judge', 'expoJudgingSession'] })
 export class ExpoJudgingSessionContext extends Node<ExpoJudgingSessionContext> {
-  @OneToOne({ entity: () => Judge, ref: true, unique: false })
+  @ManyToOne({ entity: () => Judge, ref: true, unique: false })
```
Ansaar-Syed, line 13: shouldn't it be many to one since one judge can have multiple expo judging session?

SpencerKaiser: Yep, that's why it's being changed to `ManyToOne`; `OneToOne` was wrong

Ansaar-Syed: my bad i read that as Many to Many idk why :/

Ansaar-Syed (approved):

PR #553, opened 2023-10-04
fix: schedule scroll behavior

SpencerKaiser:
## Pre-Requisites


packages/web/src/components/EventsList/EventCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <Text>🕑</Text>
+      </Flex>
+    </Tag>
+  );
+
+export const EventCard: React.FC<EventCardProps> = ({ event, containerRef }) => {
+  const { name, description, start, end } = event;
+  const inProgress = dayjs().isBetween(start, end);
+
+  const format = 'h:mm a';
```
Ansaar-Syed, line 41: does the h here mean: `Represents the hours without leading zeros`?

SpencerKaiser: That's exactly what it means!

PR #556, opened 2023-10-11
feat: Results get endpoint

Ansaar-Syed:
## Pre-Requisites


packages/api/src/api/expoJudgingSession/index.ts
```diff
@@ -10,3 +10,4 @@ export const expoJudgingSession = Router();
 expoJudgingSession.post('', adminMiddleware, post);
 expoJudgingSession.get('', adminMiddleware, get);
 expoJudgingSession.use('/:id', judgeMiddleware, id);
+expoJudgingSession.use('/:id', adminMiddleware, id);
```
charliealbright, line 13: What's the need for this?

SpencerKaiser: The only folks that have access to results should be admins so this adds admin middleware to the route... however, this is not the right way to do it...

@Ansaar-Syed you'll need to remove `judgeMiddleware` from the line above this and have the route itself manage it's middleware one level higher up the hierarchy

Ansaar-Syed: done

packages/api/src/api/expoJudgingSession/id/results/index.ts
```diff
@@ -0,0 +1,6 @@
+import { Router } from 'express';
+import { get } from './get';
+
+export const results = Router();
```
SpencerKaiser, line 4: Here's how I know you didn't test this method 😉 
```suggestion
export const results = Router({ mergeParams: true });
```

packages/api/tests/api/expoJudgingSession/index.test.ts
```diff
@@ -54,15 +52,15 @@ describe('/expoJudgingSession post endpoint registration', () => {
     });
   });
 
-  it('uses adminMiddleware and registers the GET route for the handler', async () => {
-    await jest.isolateModulesAsync(async () => {
-      const { expoJudgingSession } = await import('../../../src/api/expoJudgingSession');
+  // it('uses adminMiddleware and registers the GET route for the handler', async () => {
```
SpencerKaiser, line 55: Delete this if it's no longer being used

SpencerKaiser (changes requested):
Super close, just a few changes

SpencerKaiser (approved):

PR #560, opened 2023-10-16
Results Page

alexanderallen-aa:
## Pre-Requisites


packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  ExpoJudgingSessionResult,
+  fetchExpoJudgingSessionResults,
+} from '../../../../pageUtils/expoJudgingSession/[id]/fetchExpoJudgingSessionResults';
+import { openErrorToast } from '../../../../components/utils/CustomToast';
+
+const ExpoJudgingSessionDetails: NextPage = () => {
+  const { expoJudgingSession } = useExpoJudgingSessionFetcher();
+  const [expoJudgingSessionResults, setExpoJudgingSessionResults] =
+    React.useState<ExpoJudgingSessionResult[]>();
+  React.useEffect(() => {
```
Comment (SpencerKaiser) [inline packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx:16]:
```suggestion

  React.useEffect(() => {
```

packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          return;
+        }
+        if (Array.isArray(results)) {
+          if (results) {
+            if (results.length === 0) {
+              openErrorToast({ title: 'No results found' });
+            }
+            setExpoJudgingSessionResults(results);
+          }
+        }
```
Comment (SpencerKaiser) [inline packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx:43]: The type checking here should be exhaustive so the only other potential type for this is an array so these should be unnecessary
```suggestion


        if (results.length === 0) {
          openErrorToast({ title: 'No results found' });
        }
        setExpoJudgingSessionResults(results);

```

packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    };
+    if (expoJudgingSession) {
+      void getExpoJudgingSessionResults();
+    }
+  }, [expoJudgingSession]);
+
+  return (
+    <PageContainer
+      pageTitle={'Expo Judging'}
+      heading={'Expo Judging'}
```
Comment (SpencerKaiser) [inline packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx:54]: I think this is a better title; thoughts?
```suggestion
      pageTitle={'Expo Judging Overview'}
      heading={'Expo Judging Overview'}
```

SpencerKaiser (changes requested):
Small tweaks and then good to go

packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx
Comment (alexanderallen-aa) [reply packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx:16]: that's not me, that is prettier.

Comment (SpencerKaiser) [reply packages/web/src/pages/admin/expoJudgingSession/[id]/index.tsx:17]: I merged the commit from the UI and Prettier is fine with it

SpencerKaiser (approved):

PR #567, opened 2023-10-18
feat: Project get endpoint

Ansaar-Syed:
## Pre-Requisites

### DIFF
--- a/packages/api/src/api/project/id/get.ts
+++ b/packages/api/src/api/project/id/get.ts
@@ -8,14 +8,14 @@ export const get = async (req: Request, res: Response) => {
     params: { id: projectId },
   } = req;
   try {
-    const project = await entityManager.find(Project, { id: projectId as string });
+    const project = await entityManager.findOne(Project, { id: projectId as string });
     if (!project) {
       res.sendStatus(404);
+      return;
     }
     res.send(project);
-    console.log('hepg', project);
   } catch (error) {
-    logger.error('Unable to fetch project details from dB', error);
+    logger.error('Unable to fetch project details', error);
     res.sendStatus(500);
   }
 };
--- a/packages/api/src/api/project/id/get.ts
+++ b/packages/api/src/api/project/id/get.ts
@@ -1,0 +1,24 @@
+import { Project } from '@hangar/database';
+import { Request, Response } from 'express';
+import { logger } from '../../../utils/logger';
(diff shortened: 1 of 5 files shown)


packages/api/src/api/project/index.ts
```diff
@@ -2,10 +2,12 @@ import { Router } from 'express';
...
 import { contributors } from './contributors';
+import { details } from './id';
 
 export const project = Router();
 
 // project.get('', /* don't mount user */, get);
 project.use(mountUserMiddleware);
 project.post('', post);
 project.use('/contributors', contributors);
+project.use('/id', details);
```
SpencerKaiser, line 13: You need the colon for express to understand that it's a variable instead of just `id` as the path
```suggestion
project.use('/:id', details);
```

packages/api/src/api/project/id/index.ts
```diff
@@ -0,0 +1,6 @@
+import { Router } from 'express';
+import { get } from './get';
+
+export const details = Router({ mergeParams: true });
```
SpencerKaiser, line 4:
```suggestion
export const id = Router({ mergeParams: true });
```

Ansaar-Syed: i was thinking the same but i just felt details would be better, but i see the idea is to keep the directory name and router name the same just as Charlie also told me.

packages/api/src/api/project/id/get.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const get = async (req: Request, res: Response) => {
+  const {
+    entityManager,
+    params: { id: projectId },
+  } = req;
+  try {
+    const project = await entityManager.find(Project, { id: projectId as string });
+    if (!project) {
+      res.sendStatus(404);
```
SpencerKaiser, line 13: Missing a return after this; it'll fall through and you'll get an error like `cannot set headers after they have been set`

packages/api/src/api/project/id/get.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+    res.send(project);
+    console.log('hepg', project);
+  } catch (error) {
+    logger.error('Unable to fetch project details from dB', error);
```
SpencerKaiser, line 18:
```suggestion
    logger.error('Unable to fetch project details', error);
```

SpencerKaiser (approved):

PR #603, opened 2023-11-09
feat: Edit Project Put Schema

Ansaar-Syed:
## Pre-Requisites


packages/shared/src/schema/project/id/put.ts
```diff
@@ -0,0 +1,27 @@
...
+  location: z.string().trim().max(PutValidation.MAX_LOCATION_LENGTH).optional(),
+  repoUrl: z
+    .string()
+    .trim()
+    .url()
+    .refine(
+      (url) => url.startsWith('https://github.com/') || url.startsWith('https://gitlabs.com/'),
+      'Not a supported repo hosting platform',
+    ),
+});
```
SpencerKaiser, line 27: This is all the exact same schema from the `post` endpoint... do you think there's a way to move it out so we don't have repeated code?

packages/api/src/api/project/id/put.ts
SpencerKaiser, line 1: Since this PR doesn't implement the API endpoint, let's remove this

packages/shared/src/schema/project/id/put.ts
SpencerKaiser, line 1: I see what you're going for with the path here, but let's keep it consistent with the others and do the following:
```
packages/shared/src/schema/project
  post.ts
  put.ts
  index.ts // Exports both
```

SpencerKaiser (changes requested):

packages/api/src/api/project/id/put.ts
Ansaar-Syed: Sure thing

packages/shared/src/schema/project/id/put.ts
Ansaar-Syed: Thinking of maybe using the .merge function offered by zod

SpencerKaiser (approved):

PR #625, opened 2024-01-03
feat: Add hook to handle invite code

alexanderallen-aa:


packages/web/src/components/useProjectInviteCodeHandler/useProjectInviteCodeHandler.tsx
```diff
@@ -0,0 +1,60 @@
...
+      try {
+        project = (await axios.put(`/api/project/contributors`, { inviteCode: projectInviteCode }))
+          .data;
+      } catch {
+        toasts.Invalid();
+        return;
+      }
+      toasts.Success();
+
+      await useUserStore.getState().fetchUser();
```
SpencerKaiser, line 54: I don't think we need to wait on this; do you?
```suggestion
      void useUserStore.getState().fetchUser();
```

SpencerKaiser (commented):
One teensy suggestion but otherwise it looks great!

SpencerKaiser (approved):

